PR #13, opened 2020-03-18
Add implementation on of sys/time.h

Coquinho:
This should add all needed functions/structures from sys/time.h on a Windows fashion.

### DIFF
--- a/src/lib/evil/unposix/sys/time.h
+++ b/src/lib/evil/unposix/sys/time.h
@@ -2,7 +2,7 @@
 #define SYS_TIME_H
 
 #ifndef WIN32_LEAN_AND_MEAN
-# define WIN32_LEAN_AND_MEAN 1
+# define WIN32_LEAN_AND_MEAN
 #endif
 // Windows Kit for Windows 10 already defines `struct timeval` and `time_t`
 #include <winsock2.h>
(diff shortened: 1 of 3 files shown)


caiosignor:
looks good

src/lib/evil/unposix/sys/time.h
```diff
@@ -1,26 +1,43 @@
-#ifndef TIME_H
-#define TIME_H
-
-#include "unimplemented.h"
+#ifndef SYS_TIME_H
+#define SYS_TIME_H
 
+#ifndef WIN32_LEAN_AND_MEAN
+# define WIN32_LEAN_AND_MEAN 1
```
caiosignor: what is this ?

Coquinho: Window.h automatically includes a lot of other headers, using WIN32_LEAN_AND_MEAN it only includes what is explicitly included. I think that, for safety, everywhere you include windows.h you should verify if WIN32_LEAN_AND_MEAN is set (Am I right @JPTIZ ? ).

JPTIZ: > Window.h automatically includes a lot of other headers, using WIN32_LEAN_AND_MEAN it only includes what is explicitly included. I think that, for safety, everywhere you include windows.h you should verify if WIN32_LEAN_AND_MEAN is set (Am I right @JPTIZ ? ).

Mostly that. Not exactly only what is explicitly included, just prevents messing up trying to use two conflicting APIs at the same time (e.g. winsock and winsock2).

Coquinho:
@JPTIZ, @caiosignor there were some changes, could you have a look again?

PR #18, opened 2020-03-25
Windows:sys/types.h: Should add every needed non pthread_* type

Coquinho:
It does so by:
- including definitions that Windows already has;
- `typedef` the ones that Windows doesn't have, following `sys/types.h`
  manual guidelines;
- `pid_t` is a process id, so on this implementation is
  `typedef` as an `DWORD`, the type returned by `GetCurrentProcessId`
  function, and thus it is an `unsigned long`. `sys\types.h` manual
  establishes that it should be a `signed int` and when changed, it
  should not be larger than a `long`;

Depends on PR#16
Test Plan:
- comment `-Wno-missing-variable-declarations` at `meson.build`;
- added `-k0` to `NINJAFLAGS` at `build.bat`;
- there should not be any errors/warnings about sys/types.h.

### DIFF
--- a/src/lib/evil/unposix/sys/types.h
+++ b/src/lib/evil/unposix/sys/types.h
@@ -1,11 +1,27 @@
 #ifndef UNPOSIX_SYS_TYPES_H
 #define UNPOSIX_SYS_TYPES_H
 
-#include <stdint.h>
+#ifndef WIN32_LEAN_AND_MEAN
+# define WIN32_LEAN_AND_MEAN
+#endif
 
 #ifdef _MSC_VER
 #include <BaseTsd.h>
-#define ssize_t SSIZE_T
+typedef SSIZE_T ssize_t;
 #endif
+
+
+#include <stdint.h>
+#include <time.h>
+#include <wchar.h>
+#include <crtdefs.h>
+#include_next <sys/types.h>
+
+typedef int clockid_t;
+typedef int gid_t;
+typedef int mode_t;
+typedef unsigned long pid_t;
+typedef long suseconds_t;
+typedef int uid_t;
 
 #endif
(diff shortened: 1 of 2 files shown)


joaoantoniocardoso:
Sadly I got some errors trying to compile in the following way:

```
git checkout 0a9adef307d2372a452d77aa29b2ad2f6dd1ae3a
git pull expt devs/joaoantoniocardoso/build-on-windows-with-clang-cl/meson-for-evil-unposix
\rm -rf build sys/build && meson build > meson.log
cd build && ninja > ../ninja.log
```
[meson.log](https://github.com/expertisesolutions/efl/files/4382746/meson.log)
[ninja.log](https://github.com/expertisesolutions/efl/files/4382749/ninja.log)

src/lib/evil/unposix/sys/types.h
```diff
@@ -1,11 +1,27 @@
 #ifndef UNPOSIX_SYS_TYPES_H
 #define UNPOSIX_SYS_TYPES_H
 
-#include <stdint.h>
+#ifndef WIN32_LEAN_AND_MEAN
+# define WIN32_LEAN_AND_MEAN
+#endif
 
 #ifdef _MSC_VER
```
JPTIZ: I think it would make more sense to let this `ifndef` inside `#ifdef _MSC_VER`, but to be honest I can't think why I've previously let that `_MSC_VER`: this header exists only when compiling for Windows. Hm.

JPTIZ: Good thing I've slept. Now that I'm awake that I remembered Evil isn't just for Windows, but for MacOS too.

Coquinho: Wouldn't be batter to put all the files on a #ifdef then? like:
#ifndef UNPOSIX_<dependence>

JPTIZ: That would be better when we reach MacOS. Since we don't know what on Mac is not implemented in Windows, we don't know which missing parts do intersect.

JPTIZ:
Looks good. Gonna merge tonight.

PR #19, opened 2020-03-27
Devs/joaoantoniocardoso/locale

joaoantoniocardoso:
It did compiple, but I have some considerations:
- *unix locale* uses __THROW, so a minimal `cdefs.h` was added too.
- `efl/src/lib/eina/eina_private.h` was declaring a conflicting `locale_t` for `_WIN32`, I just commented that lines, but I don't know if there will be consequences for doing this.
- a big part of *unix locale* depends on `__USE_XOPEN2k8`, like `struct lconv` declaration, but this flag isn't set in by our build setup, so I removed that.
- compared to the *unix locale*, `CATEGORY_MASK` is missing, it is not used yet, should it be added **now**?


joaoantoniocardoso:
The idea here was to request a review for @caiosignor, testing against his branch. We could test this using patches or a rebase, a PR is not necessary.
Closing until locale is fully implemented.

PR #35, opened 2020-04-16
Windows: Add getopt implementation

Coquinho:
Summary:
========
Add getopt from https://github.com/nanoporetech/getopt-win32 but with some
modifications for better suite our purpose, like removing wchar
implementation;

- `src/lib/evil/unposix/getopt.c` implements getopt, getopt_long and
  getopt_long_only - EFL only use the first and second, but I'm unsure if
  should remove the latter;

- `src/tests/evil/unposix/evil_unposix_test_getopt.c` implements a little teste
  case for getopt.c - it is not a unit tests case;

- For now getopt.h should be included as `#include "getopt.h"` - i.e
  with quotes, not brakets;

Test Plan:
==========
For testing the getopt:
- As everything that evil_unposix_test_getopt.c, getopt.{c,h} are unplemented at
  unposix (for now at least) they should include Windows CRT versions and not
  unposix ones.
- With evil_unposix_test_getopt.c and getopt.{c,h} at current dir:
```
   clang-cl evil_unposix_test_getopt.c getopt.c -DSTATIC_GETOPT
```
  Where `-DSTATIC_GETOPT` export getopt.{c,h} variables and functions for static
  building;
- Now you can test getopt by passing args to the exe, e.g:
```
  evil_unposix_test_getopt.exe.exe -0 -1 -2 --create banana
```
  Where the possible short and long options are defined at
  evil_unposix_getopt.c

For testing if everything else is ok:
- Comment `Wno-implicit-function-declaration` at meson.build;
- Run `configure.bat`;
- Run `build -k0`, it should only return error/warnigs that
  `devs/expertise/native-windows` does

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -175,7 +175,9 @@ if sys_windows == true
       # ------------------------------------
       # Default flags for native compilation
       '-Wno-language-extension-token',
-      '-DWIN32_LEAN_AND_MEAN'
+      '-DWIN32_LEAN_AND_MEAN',
+      '-DSTATIC_GETOPT',
+
     ], language : 'c')
 endif
 
--- a/src/lib/evil/unposix/getopt.h
+++ b/src/lib/evil/unposix/getopt.h
@@ -1,28 +1,111 @@
-#ifndef UNPOSIX_GETOPT_H
-#define UNPOSIX_GETOPT_H
+/* Getopt for Microsoft C
+This code is a modification of the Free Software Foundation, Inc.
+Getopt library for parsing command line argument the purpose was
+to provide a Microsoft Visual C friendly derivative. This code
+provides functionality for both Unicode and Multibyte builds.
+Date: 02/03/2011 - Ludvik Jerabek - Initial Release
+Version: 1.0
+Comment: Supports getopt, getopt_long, and getopt_long_only
+and POSIXLY_CORRECT environment flag
+License: LGPL
+Revisions:
+02/03/2011 - Ludvik Jerabek - Initial Release
+02/20/2011 - Ludvik Jerabek - Fixed compiler warnings at Level 4
+07/05/2011 - Ludvik Jerabek - Added no_argument, required_argument, optional_argument defs
+08/03/2011 - Ludvik Jerabek - Fixed non-argument runtime bug which caused runtime exception
+08/09/2011 - Ludvik Jerabek - Added code to export functions for DLL and LIB
+02/15/2012 - Ludvik Jerabek - Fixed _GETOPT_THROW definition missing in implementation file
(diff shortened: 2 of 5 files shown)


joaoantoniocardoso:
To simplify testing you could script a `src/tests/evil/unposix/test.sh` like this:
```
# update
cp ../../../../src/lib/evil/unposix/getopt.[ch] .
# build
clang-cl evil_unposix_test_getopt.c getopt.c -I . -DSTATIC_GETOPT
# test
./evil_unposix_test_getopt.exe -0 -1 -2 --create banana
```

meson.build
```diff
@@ -175,7 +175,9 @@ if sys_windows == true
       # ------------------------------------
       # Default flags for native compilation
       '-Wno-language-extension-token',
-      '-DWIN32_LEAN_AND_MEAN'
+      '-DWIN32_LEAN_AND_MEAN',
+      '-DSTATIC_GETOPT',
+
```
caiosignor, line 180: i think you can remove this new line

src/lib/evil/unposix/getopt.h
```diff
@@ -1,28 +1,111 @@
...
+PURPOSE, OR NON-INFRINGEMENT. SOME JURISDICTIONS DO NOT ALLOW THE
+EXCLUSION OF IMPLIED WARRANTIES, SO THE ABOVE EXCLUSION MAY NOT
+APPLY TO YOU. IN NO EVENT WILL I BE LIABLE TO ANY PARTY FOR ANY
+DIRECT, INDIRECT, SPECIAL OR OTHER CONSEQUENTIAL DAMAGES FOR ANY
+USE OF THIS MATERIAL INCLUDING, WITHOUT LIMITATION, ANY LOST
+PROFITS, BUSINESS INTERRUPTION, LOSS OF PROGRAMS OR OTHER DATA ON
+YOUR INFORMATION HANDLING SYSTEM OR OTHERWISE, EVEN If WE ARE
+EXPRESSLY ADVISED OF THE POSSIBILITY OF SUCH DAMAGES.
+*/
+#ifndef __GETOPT_H_
```
caiosignor, line 34: identantion here?

Coquinho: This PR isn't respecting the EFL code conventions for now, because I'm not sure it will make to `native-windows` as @JPTIZ is working in adding it as a dependency.

Coquinho:
Should I delete this PR already? Or wait for #45 to be landed?

JPTIZ:
> Should I delete this PR already? Or wait for #45 to be landed?

I think we can close this, yeah.

Issue #30, opened 2020-04-02
Implement missing fnmatch function


JPTIZ:
```c
int fnmatch(const char * a, const char * b, int c)
```

vtorri:
regex and fnmath can be get from musl : 

https://github.com/vtorri/ewpi/blob/master/packages/regex/

JPTIZ:
For regex, we used `pcre` (with a Wrap-Dependency straight from WrapDB: https://wrapdb.mesonbuild.com/pcre).

In order to use `fnmatch` from musl, would it require to replace that pcre dependency? Or we can pick musl's fnmatch implementation stand-alone?

vtorri:
musl also provides regex stuff, so fnmatch + regex with few c files. so i would say, replace pcre with what i have done. e

JPTIZ:
Okay, gonna check it then.

PR #36, opened 2020-04-16
Devs/joaoantoniocardoso/fnmatch

joaoantoniocardoso:
Now a working fnmatch using regular expressions from `pcre`, aiming to solve #30 .

The idea here is a function that converts wildcard/glob patterns to regular expressions (with regards to the special flags passed to the original posix fnmatch) and using regex checks if it matches, furthermore it intended to work with both unix-like and windows-like paths.

Why this strategy? Because there was no equivalent function that could handle globs for both windows-like and unix-like paths, `pcre` is already available and using the glibc fnmatch would require a great amount of work too. Also maybe all fnmatch use cases in efl could be using regex instead of fnmatch...

Something that I ignored was that if the string passed to fnmatch ends with '/' or '\\' and the glob don't specify an ending match (say '/' or '\\' if using `FNM_PATHNAME`), than it will not result in a match, and maybe is not the expected behavior.

Another point is that I was unable to compile this using the current configuration because meson.build is not correctly linking to pcre and therefore it doesn't find `<pcreposix.h>` both in `evil/pcre/regex.h` and in this `evil/unposix/fnmatch.h`. ~~Because of that, this PR is not ready to merge~~. 

~~Note: checkout `devs/joaoantoniocardoso/fnmatch-with-tests` to see the development test files.~~

Updates: 
- ~~This branch was rebased and should be merged after #37.~~ rebased to devs/expertise/native-windows again after #37 merged.
- Tested with pcre on windows (using wrap-dependency branch)
- Tested on linux (for development) and windows (The development test files was moved to `src/tests/evil/unposix`.)


Coquinho:
So, about the PCRE, your idea is to create a new PR that handles that?

src/lib/evil/unposix/fnmatch.c
```diff
@@ -0,0 +1,271 @@
...
+                      if (pattern[j] == '*')
+                        {
+                           if (j == 0)
+                             {
+                                strcpy(reg_pattern + i, "[^\\.][^/]*");
+                                i += 10;
+                             }
+                           else
+                             {
+                                if (pattern[j - 1] == '/')
```
Coquinho, line 133: Not exactly asking a change, but why you have so much:
```
if () {}
else 
{
  if () {}
  else {}
}
```
Why not making those `else if`'s?
```
if () {}
else if () {}
else {}
```

joaoantoniocardoso: Yeah, that was the way my mind deal with the decision tree and of course it could be rewritten for clarity. Do you think that make it more flat will be easier to understand? I will see how it looks when flatted...

Coquinho: For me at least `else if`'s makes it easier for understanding the flow.

joaoantoniocardoso:
Tested with wrap-dependency for pcre #37 and it builds :)

src/lib/eina/_eina_fnmatch.c
```diff
@@ -0,0 +1,256 @@
...
+            "$\0");
+
+   return reg_pattern;
+}
+
+int
+fnmatch(const char *pattern, const char *string, int flags)
+{
+   regex_t regex;
+   int result;
```
JPTIZ, line 224: I suggest moving this to below in `int result = regexec(...)`.

src/lib/eina/_eina_fnmatch.c
```diff
@@ -0,0 +1,256 @@
...
+            sizeof(reg_pattern + i),
+            "$\0");
+
+   return reg_pattern;
+}
+
+int
+fnmatch(const char *pattern, const char *string, int flags)
+{
+   regex_t regex;
```
JPTIZ, line 223: I suggest moving this to below, above `if (regcomp(&regex...))`. It would point out better when is it initialized.

src/lib/eina/eina_fnmatch.h
```diff
@@ -0,0 +1,33 @@
...
+#ifdef _WIN32
+
+/* Bits set in the FLAGS argument to `fnmatch'.  */
+#define FNM_PATHNAME (1 << 0) /* No wildcard can ever match `/'.  */
+#define FNM_NOESCAPE (1 << 1) /* Backslashes don't quote special chars.  */
+#define FNM_PERIOD   (1 << 2) /* Leading `.' is matched only explicitly.  */
+
+#define FNM_FILE_NAME   FNM_PATHNAME /* Preferred GNU name.  */
+#define FNM_LEADING_DIR (1 << 3)     /* Ignore `/...' after a match.  */
+#define FNM_CASEFOLD    (1 << 4)     /* Compare without regard to case.  */
```
JPTIZ, line 18: (Not a change request)

About those "table statements" I mentioned earlier: in this case (sequences of `#define`) it's pretty fine.

src/lib/eina/eina_fnmatch.h
```diff
@@ -0,0 +1,33 @@
...
+
+/* Value returned by `fnmatch' if STRING does not match PATTERN.  */
+#define FNM_NOMATCH 1
+
+int fnmatch(const char *pattern, const char *string, int flags);
```
JPTIZ, line 23: In Eina, this would be called `eina_fnmatch`.

src/lib/eina/meson.build
```diff
@@ -199,6 +200,7 @@ sources = [
 
 if sys_windows == true
   sources += 'eina_file_win32.c'
+  sources += '_eina_fnmatch.c'
```
JPTIZ, line 203: I know we talked about this earlier, but anyway: you should call it `eina_fnmatch.c` :) The `_` is just for private things in source-code, not filenames.

Coquinho, line 203: Another thing here, this way `eina_fnmatch` will not be available to all builds as intended, only to a Windows one.

joaoantoniocardoso: Is it intended to be available to all platforms? I was thinking that others like linux will continue to use fnmatch from gnu fnmatch.h, so instead compile eina_fnmatch.c, it would `#define eina_fnmatch fnmatch`

Coquinho: hmmmm, ok I think you really can do that. So your idea is to add a `#define eina_fnmatch fnmatch` after `#endif /* _WIN32 */` at `eina_fnmatch.h`?

Coquinho: Oh, you have already done it. Yay

JPTIZ:
Would you mind moving `eina_fnmatch` to `evil_fnmatch.{c,h}` (and renaming to `fnmatch`)?

Issue #38, opened 2020-04-20
Compile devs/expertise/native-windows on Linux


Coquinho:
Our `devs/expertise/native-windows` should bring a native windows alternatives
to compile EFL, but it shouldn't interfere with the Linux version.

For now, `devs/expertise/native-windows` is broke when compiling on Linux.

PR #39, opened 2020-04-20
Compile `devs/expertise/native-windows on Linux

Coquinho:
Bug: Compile `native-windows` on Linux

Summary:
========
This PR makes `devs/expertise/native-windows` compile on linux.

lua was a little problematic:
luajit is a default option to bindings in `master` but at
`devs/expertise/native-windows` it was changed to lua, This PR brings
luajit back as a default option, but make a Windows build always use
lua - semantically equal to what happened at
`devs/expertise/native-windows`

Ref Issue #38

Linux Test plan:
================
- `meson` configured with `-Dbindigns=cxx,mono`;
- `ninja` should return every warning that `master` does, nothing else;
- `meson test -C <build dir>` should pass in everything that `master`
  does;

Windows Test plan:
==================
- Run `configure.bat` normally;
- Run `build.bat` it should yield the same erros/warnigs that
  `devs/expertise/native-windows` does;

### DIFF
--- a/src/bindings/cxx/meson.build
+++ b/src/bindings/cxx/meson.build
@@ -9,8 +9,14 @@ cxx_sublibs = [
   ['Eet',         true,  true, false, []],
   ['Eio',        false, false,  true, []],
   ['Evas',       false,  true,  true, []],
-  ['Edje',       false, false,  true, []],
-#  ['Eldbus',      true,  true,  true, []],
+  ['Edje',       false, false,  true, []],]
+
+if (not sys_windows)
+  cxx_sublibs += [['Eldbus',      true,  true,  true, []]]
+  assert(cxx_sublibs[-1] == ['Eldbus',      true,  true,  true, []])
+endif
+
+cxx_sublibs += [
   ['Elementary',  true,  true,  true, []]
 ]
 
--- a/src/bindings/cxx/meson.build
+++ b/src/bindings/cxx/meson.build
@@ -11,9 +11,8 @@ cxx_sublibs = [
   ['Evas',       false,  true,  true, []],
   ['Edje',       false, false,  true, []],]
 
-if (not sys_windows)
+if not sys_windows
   cxx_sublibs += [['Eldbus',      true,  true,  true, []]]
-  assert(cxx_sublibs[-1] == ['Eldbus',      true,  true,  true, []])
 endif
 
 cxx_sublibs += [
--- a/src/bindings/meson.build
+++ b/src/bindings/meson.build
@@ -1,6 +1,6 @@
 
 bindings = get_option('bindings')
-if (not sys_windows)
+if not sys_windows
   bindings_order = ['luajit', 'cxx', 'mono']
 else
   bindings_order = ['lua', 'cxx', 'mono']
--- a/src/bindings/meson.build
+++ b/src/bindings/meson.build
@@ -1,6 +1,10 @@
 
 bindings = get_option('bindings')
-bindings_order = ['cxx', 'mono']
+if (not sys_windows)
+  bindings_order = ['luajit', 'cxx', 'mono']
+else
+  bindings_order = ['lua', 'cxx', 'mono']
+endif
 
 if (get_option('dotnet') and not bindings.contains('mono'))
   message('dotnet support requires the C# bindings')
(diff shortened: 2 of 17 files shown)


Coquinho:
I still have to remove some commits from `master`, so it still as a draft

src/bindings/cxx/meson.build
```diff
@@ -9,8 +9,14 @@ cxx_sublibs = [
   ['Eet',         true,  true, false, []],
   ['Eio',        false, false,  true, []],
   ['Evas',       false,  true,  true, []],
-  ['Edje',       false, false,  true, []],
-#  ['Eldbus',      true,  true,  true, []],
+  ['Edje',       false, false,  true, []],]
+
+if (not sys_windows)
```
JPTIZ, line 14: (This (is (not (lisp)))

src/bindings/meson.build
```diff
@@ -1,6 +1,10 @@
 
 bindings = get_option('bindings')
-bindings_order = ['cxx', 'mono']
+if (not sys_windows)
```
JPTIZ, line 3: > (This (is (not (lisp)))

src/generic/evas/meson.build
```diff
@@ -9,10 +9,22 @@ generic_deps = []
 
 subdir('common')
 
+inc_dir = config_dir
+common_deps = generic_deps
+if (sys_windows)
```
JPTIZ, line 14: > > (This (is (not (lisp)))

src/generic/evas/meson.build
```diff
@@ -9,10 +9,22 @@ generic_deps = []
...
+  inc_dir = [inc_dir, zlib_include_dir]
+  assert(inc_dir[0] == config_dir)
+  assert(inc_dir[1] == zlib_include_dir)
+else
+  common_deps = [generic_deps, rt]
+  assert(common_deps[0] == generic_deps)
+  assert(common_deps[1] == rt)
```
JPTIZ, line 21: What's the point behind these asserts? They sound like testing meson features.

src/generic/evas/xcf/meson.build
```diff
@@ -4,5 +4,9 @@ generic_src = files([
   'pixelfuncs.c'
 ])
 
-generic_deps = [eina, zlib]
+if (not sys_windows)
```
JPTIZ, line 7: > > > (This (is (not (lisp)))

src/lib/ecore/meson.build
```diff
@@ -196,10 +196,17 @@ if get_option('systemd') == true
   ecore_deps += systemd
 endif
 
+if (not sys_windows)
```
JPTIZ, line 199: > > > > (This (is (not (lisp)))

src/lib/eet/meson.build
```diff
@@ -1,4 +1,9 @@
 eet_deps = [crypto, jpeg, rg_etc]
+if not sys_windows
+  eet_deps += m
+  assert(eet_deps[-1] == m)
```
JPTIZ, line 4: Is this assert really needed?

src/lib/evas/meson.build
```diff
@@ -4,17 +4,24 @@
...
   giflib = cc.find_library('gif')
+elif get_option('evas-loaders-disabler').contains('gif')
+  giflib = declare_dependency()
 endif
+
 webp = dependency('libwebp', required: get_option('evas-loaders-disabler').contains('webp') == false)
 libopenjp2 = dependency('libopenjp2', required: get_option('evas-loaders-disabler').contains('jp2k') == false)
 
-if get_option('evas-loaders-disabler').contains('generic')
+if (sys_windows and get_option('evas-loaders-disabler').contains('generic'))
```
JPTIZ, line 24: No need for () here, the `and` operator will be properly handled :)

src/bindings/meson.build
Coquinho: if (not (sys_windows))

src/lib/eet/meson.build
Coquinho: it was more like a to see if I was working correctly with the meson Array, I can remove all the new ones if you think it is better

PR #42, opened 2020-04-24
windows: Add strings.h functions to work with ucrt on windows

caiosignor:
# Goal
This aims to solve strings.h dependency from posix, to work with native Universal C Runtime of windows.

# How it work
From posix strings.h functions, we have: 
- ffs
- strcasecmp
- strncasecmp
- strcasecmp_l
- strncasecmp_l

In case of `ffs`, i've made a dumb implementation located in 'src\lib\evil\unposix\strings.c'.

In all remaining functions, i did an #define `<posix_function>` to `_<ucrt_functions>`

# how do I test it
You can configure and build this branch on windows with: .\configure && .\build -k0, there isn't should be errors, with related functions.

I made too a .c code that use theses functions. These code are located in src\tests\evil\unposix\test_evil_unposix_strings.c. 
To run it, you need execute ./run_test.sh int git bash.


src/tests/evil/unposix/test_evil_unposix_strings.c
```diff
@@ -0,0 +1,46 @@
...
+#include "strings.h"
+#include <stdio.h>
+
+int equals_strings(){
+    char *str1,*str2;
+    str1 = "fOo";
+    str2 = "FoO";
+    
+    return strcasecmp(str1,str2);
+}
```
JPTIZ, line 11: Why not...
```suggestion
int equals_strings()
{
   return strcasecmp("fOo", "FoO");
}
```

It even has a plus of fitting EFL coding conventions :)

src/lib/evil/unposix/strings.c
```diff
@@ -0,0 +1,10 @@
+#include "strings.h"
+
+inline int
```
JPTIZ, line 3: No `inline` in this case. It just makes sense if it were defined in a header.

src/tests/evil/unposix/test_evil_unposix_strings.c
```diff
@@ -0,0 +1,46 @@
...
+
+int str1_n_equals_str2(){
+    char *str1,*str2;
+    str1 = "fOoO";
+    str2 = "fOo";
+    
+    return strncasecmp(str1,str2,3);
+}
```
JPTIZ, line 35: Ditto.

src/tests/evil/unposix/test_evil_unposix_strings.c
```diff
@@ -0,0 +1,46 @@
...
+int main()
```
JPTIZ, line 36:
```suggestion

int main()
```

src/tests/evil/unposix/test_evil_unposix_strings.c
caiosignor: loll, i'll change it

joaoantoniocardoso: According to EFL style..:

```suggestion
int 
equals_strings()
{
    return strcasecmp("fOo", "FoO");
}
```

src/lib/evil/unposix/strings.c
caiosignor: i'm confused, it is defined into strings.h

src/benchmarks/eina/eina_bench.c
```diff
@@ -111,7 +111,7 @@ main(int argc, char **argv)
       case 2:
          for (i = 0; etc[i].bench_case; ++i)
            {
-              if (strcasecmp(etc[i].bench_case, argv[1]))
+              if (evil_strcasecmp(etc[i].bench_case, argv[1]))
```
JPTIZ, line 114: You do not use `evil_*` explicitly. The idea of Evil is that noone knows about Evil despiste including it as a header.

src/lib/evil/evil_strings.h
```diff
@@ -0,0 +1,21 @@
...
+# include <Windows.h>
+# undef WIN32_LEAN_AND_MEAN
+#else
+# include<string.h>
+#endif
+
+int evil_strncasecmp(const char *string1, const char *string2, size_t count);
+int evil_strcasecmp(const char *string1, const char *string2);
+int evil_strncasecmp_l(const char *string1, const char *string2, size_t count, _locale_t locale);
+int evil_strcasecmp_l(const char *string1, const char *string2, _locale_t locale);
```
JPTIZ, line 17: Inside Evil, call them just `strncasecmp`, `strcasecmp`, etc. (without `evil_*`).

src/lib/eina/eina_simple_xml_parser.c
```diff
@@ -21,9 +21,8 @@
 # include "config.h"
 #endif
 
-#include <strings.h>
 #include <stdlib.h>
-#include <string.h>
+#include "evil_string.h"
```
JPTIZ, line 25: Forgot to tell, but `evil_*` are never included this way. You always include `<evil_private.h>`.

PR #43, opened 2020-04-25
Fix DLL linkage

JPTIZ:
It should stop at missing `pthread_cleanup_{push,pop}` symbols in linkage.

- Adds `<SUBLIB_NAME>_API` in order to properly set `__declspec(dllimport/dllexport)` for public APIs for the following libraries: Eo, Eolian, Eina, Evil, Unposix, Edje and Elementary;
- Fix my previous mistakes when including `evil_*.h`.

Since most of these 940 files are changes from `EAPI` to `<something>_API`, code-review may be done just in `evil_*` files and in unposix. `meson.build` would be a bonus but not necessary.


JPTIZ:
I think this is ready for review __for now__.

src/lib/evil/evil_fcntl.h
```diff
@@ -107,7 +107,7 @@ struct flock
  *
  * @ingroup Evil
  */
-EAPI int fcntl(int fd, int cmd, ...);
+//EVIL_API int fcntl(int fd, int cmd, ...);
```
joaoantoniocardoso, line 110: Is it right?

JPTIZ: Not at all, oops!

felipealmeida:
Why EAPI have been removed without being replaced with anything else in some libraries (like ecore.c) ? This doesn't make much sense.

JPTIZ:
> Why EAPI have been removed without being replaced with anything else in some libraries (like ecore.c) ? This doesn't make much sense.

AFAIU, `__declspec` is only needed in declaration (and so only in header), isn't it? Or should them go in the .c file too (in order to check consistency)?

felipealmeida:
Needed or not, the commit should be as unintrusive as possible. We should probably break the patch down in libraries so we can test the changes.

felipealmeida:
Unfortunately , there's a *lot* of unneeded changes like code style changes around evil, etc.

felipealmeida:
I'm doing a cleanup in devs/felipealmeida/fix-eina-dll-export for this PR. Let me assume this PR and organize it.

Issue #25, opened 2020-04-02
Implement missing stdlib functions


JPTIZ:
The following must be implemented:

```c
int mkstemp(char* template);
int mkostemp(char* template, int flags);
int mkostemps(char* template, int suffixlen, int flags);
```

vtorri:
why should they be implemented ? mko* functions are not used in the EFL

JPTIZ:
> why should they be implemented ? mko* functions are not used in the EFL

These issues were opened as compile errors emerged (of missing headers or implicit function declaration). Some of them I just checked which functions that header implemented, should've seen this before, since indeed `mkostemps` is not used at all.

For `mkostemp`, I used `git grep -n "mkostemp"` and it returned:

`src/modules/ecore_buffer/shm/ecore_buffer_shm.c:73:   fd = mkostemp(name, O_CLOEXEC);`

Should that module be skipped in Windows?

vtorri:
tell me ! ecore_buffer is for wayland and is not compiled at all on Windows, so these functions are not used

PR #48, opened 2020-04-29
Implement missing stdlib functions mkstemp, mkostemp and mkostemps

joaoantoniocardoso:
To solve #25, this PR should implement `mkstemp`, `mkstemps`, `mkostemp` and `mkostemps` to be built under windows environment.

For now just the header and a initial test was created.


joaoantoniocardoso:
Because #57 seems a better approach to this problem, this PR is now closed.

PR #51, opened 2020-04-30
bug: efl includes <pthread.h>

caiosignor:
Efl still includes pthreads.h when it should be #include <Einha.h> outside eina/ and #include "eina_threads.h" inside eina/.

With this PR, efl will include <Eina.h> or "eina_threads.h"

### DIFF
--- a/src/examples/elementary/efl_thread_1.c
+++ b/src/examples/elementary/efl_thread_1.c
@@ -1,7 +1,7 @@
 //Compile with:
 //gcc -o efl_thread_1 efl_thread_1.c -g `pkg-config --cflags --libs elementary`
 #include <Elementary.h>
-#include <pthread.h>
+#include <Eina.h>
 
 static Evas_Object *win = NULL;
 static Evas_Object *rect = NULL;
--- a/src/lib/ecore/ecore_signal.c
+++ b/src/lib/ecore/ecore_signal.c
@@ -12,7 +12,7 @@
 #include <assert.h>
 #include <fcntl.h>
 #include <errno.h>
-#include <pthread.h>
+#include <Eina.h>
 
 #include "Ecore.h"
 #include "ecore_private.h"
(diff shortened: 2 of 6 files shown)


src/lib/eina/eina_sched_posix.c
```diff
@@ -21,7 +21,7 @@
 #include <sys/time.h>
 #include <sys/resource.h>
 #include <errno.h>
-#include <pthread.h>
+
```
joaoantoniocardoso, line 24: It shouldn't include "eina_thread.h" too?

src/examples/elementary/efl_thread_1.c
```diff
@@ -1,7 +1,7 @@
 //Compile with:
 //gcc -o efl_thread_1 efl_thread_1.c -g `pkg-config --cflags --libs elementary`
 #include <Elementary.h>
-#include <pthread.h>
+#include <eina_thread.h>
```
joaoantoniocardoso, line 4: I think that if it is not inside eina, it should include <Eina.h> and not directly <eina_thread.h>.

src/lib/eina/eina_sched_posix.c
caiosignor: it include eina_sched.h, who includes eina_thread.h

src/examples/elementary/efl_thread_1.c
caiosignor: oh

src/lib/eina/eina_sched_posix.c
joaoantoniocardoso: Hmm ok. For now it is not clear to me if it is a good practice. Personally I prefer to explicitly include all the needed headers even if it seems redundant..

JPTIZ: To be honest, personally I would prefer to include wanted dependencies explicitly, but no big deal for now.

PR #52, opened 2020-04-30
bug: efl using pthread_functions insteod of eina_thread

caiosignor:
This PR aims to solve: efl still using pthread_functions functions instead of eina-thread in  src/lib/eina/eina_debug.c.

I've changed the function that use pthread to work with eina_thread lib.

### DIFF
--- a/src/lib/eina/eina_debug.c
+++ b/src/lib/eina/eina_debug.c
@@ -597,37 +597,15 @@ _monitor(void *_data)
 static void
 _thread_start(Eina_Debug_Session *session)
 {
-#ifndef _WIN32
-   pthread_t monitor_thread;
-   int err;
-   sigset_t oldset, newset;
+   Eina_Thread monitor_thread;
 
-   sigemptyset(&newset);
-   sigaddset(&newset, SIGPIPE);
-   sigaddset(&newset, SIGALRM);
-   sigaddset(&newset, SIGCHLD);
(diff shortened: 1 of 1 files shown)


src/lib/eina/eina_debug.c
```diff
@@ -597,37 +597,15 @@ _monitor(void *_data)
...
-#endif
-   pthread_sigmask(SIG_BLOCK, &newset, &oldset);
+   Eina_Thread monitor_thread;
 
-   err = pthread_create(&monitor_thread, NULL, _monitor, session);
+   Eina_Bool err = eina_thread_create(&monitor_thread,EINA_THREAD_BACKGROUND,-1,(Eina_Thread_Cb)_monitor,session);
 
-   pthread_sigmask(SIG_SETMASK, &oldset, NULL);
-   if (err != 0)
+   if (!err)
```
joaoantoniocardoso, line 604: The return values of `pthread_create()` and `eina_thread_create()` differs?
Is it intentional changing from `(err != 0)` to `(!err)`?

caiosignor: Hi @joaoantoniocardoso according to [efl thread documentation](https://www.enlightenment.org/develop/legacy/api/c/start#group__Eina__Thread__Group.html), the return value of eina_thread_create() differs from pthread_create().

if (! err) is itentionally for adapt the new code

Issue #26, opened 2020-04-02
Implement missing fcntl functions


JPTIZ:
The following functions should be implemented:

```c
int fcntl(int a, int b, ...);
int open(const char * a, int b, ...);
int openat(int a, const char * b, int c, ...);
int posix_fadvise(int a, off_t b, off_t c, int d);
int posix_fallocate(int a, off_t b, off_t c);
```

Coquinho:
EFL only uses `fcntl()` and `open()`. 
`fcntl()` is implemented by `evil_fcntl.c`, and `open()` is implemented by #53 
So I will link #53 to this issue.

PR #53, opened 2020-05-05
Windows: Wrapper to Open/Close a File descriptor

Coquinho:
This adds `open` and `close` functions to deal with file descriptors.

`open` is defined at `evil_fcntl` and `close` at `evil_unistd`.


Coquinho:
EFL only uses `fcntl()` and `open()` from `fcntl.h` functions listed by #26. 
`fcntl()` is implemented by `evil_fcntl.c`, and `open()` is implemented here. 
So this PR may close #26.

PR #57, opened 2020-05-06
Windows: Implementation for mkstemp, mkstemps, mkostemp, mkostemps

joaoantoniocardoso:
This PR may solve #25.
- `mkstemps()` was already implemented in `evil_stdlib.c`, so it was changed to accept open flags as parameters, turning it into `mkostemps()`. 
- `mkstemp`, `mkstemps` and `mkostemp` are implemented as wrappers of `mkostemps`. 
- All files that calls one of these functions now is including `evil_private.h`.


vtorri:
this should be closed as mko* functions are not used on Windows

PR #59, opened 2020-05-10
Implementation for sys/stat.h 

ricardocvel:
This PR aims to implement the function:
- fstatat ()
for Windows;
and declare the functions in stat.h:
- umask ()
- chmod ()
(using the header: <corecrt_io.h> and <errno.h>)

### DIFF
--- a/src/lib/evil/unposix/sys/stat.c
+++ b/src/lib/evil/unposix/sys/stat.c
@@ -1,0 +1,130 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+//#include <iostream>
+#include <Windows.h>
+# include <sys/types.h>
+# include <sys/stat.h>
+#include "stat.h"
+
+
+
+#define	AT_SYMLINK_NOFOLLOW	0x01
+
+
+int fstatat(int dirfd, const char *pathname, struct stat *statbuf,int flags){
+    int r_fstatat;
+    int size_str, size_str_02;
+    char* pathbuf = NULL;
+    size_t pathbuf_size = 0;
+    DWORD copied = 0;
+
+
+
+
+    if(pathname[2] == '\\'){
+        if(flags == AT_SYMLINK_NOFOLLOW){
+            r_fstatat = stat(pathname,statbuf);
+        }
+        else{ 
+            r_fstatat = stat(pathname,statbuf);
+        }
+    }
+    else{
+        do {
+
+            pathbuf_size += MAX_PATH;
+            pathbuf = realloc(pathbuf, pathbuf_size * sizeof(char));
+            copied = GetModuleFileName(NULL, pathbuf, pathbuf_size);
+        } while (copied >= pathbuf_size);
+        pathbuf_size = copied;
+        pathbuf = realloc(pathbuf, pathbuf_size * sizeof(char)); 
+        copied = GetModuleFileName(NULL, pathbuf, pathbuf_size);
+
+
+        //getlasterror
+
+        for(size_str = strlen(pathbuf); size_str >= 0; size_str -- ){
+            if(pathbuf[size_str] == '\\'){
+                pathbuf[size_str + 1] = 0;
+                break;
+            }
+        }
+
+        size_str = strlen(pathbuf) + strlen(pathname);
+        //size_str = strlen(pathname); 
+
+        // verificar a nescesidade de " / " entre as variaveis
+        char* path_complete = malloc(sizeof(char) * size_str);
+
+        path_complete = pathbuf;
+
+
+        if(flags == AT_SYMLINK_NOFOLLOW){
+            r_fstatat = stat(path_complete, statbuf);
+        }
+        else{ 
+            r_fstatat = stat(path_complete, statbuf);
+        }
+    }
+    return r_fstatat;
+}
+
+int fstatat64(int dirfd, const char *pathname, struct stat *statbuf,int flags){
+    int r_fstatat;
+    int size_str, size_str_02;
+    char* pathbuf = NULL;
+    size_t pathbuf_size = 0;
+    DWORD copied = 0;
+
+
+
+
+    if(pathname[2] == '\\'){
+        if(flags == AT_SYMLINK_NOFOLLOW){
+            r_fstatat = stat64(pathname,statbuf);
+        }
+        else{ 
+            r_fstatat = stat64(pathname,statbuf);
+        }
+    }
+    else{
+        do {
+
+            pathbuf_size += MAX_PATH;
+            pathbuf = realloc(pathbuf, pathbuf_size * sizeof(char));
+            copied = GetModuleFileName(NULL, pathbuf, pathbuf_size);
+        } while (copied >= pathbuf_size);
+        pathbuf_size = copied;
+        pathbuf = realloc(pathbuf, pathbuf_size * sizeof(char)); 
+        copied = GetModuleFileName(NULL, pathbuf, pathbuf_size);
+
+
+        //getlasterror
+
+        for(size_str = strlen(pathbuf); size_str >= 0; size_str -- ){
+            if(pathbuf[size_str] == '\\'){
+                pathbuf[size_str + 1] = 0;
+                break;
+            }
+        }
+
+        size_str = strlen(pathbuf) + strlen(pathname);
+        //size_str = strlen(pathname); 
+
+        // verificar a nescesidade de " / " entre as variaveis
+        char* path_complete = malloc(sizeof(char) * size_str);
+
+        path_complete = pathbuf;
+
+
+        if(flags == AT_SYMLINK_NOFOLLOW){
+            r_fstatat = stat64(path_complete, statbuf);
+        }
+        else{ 
+            r_fstatat = stat64(path_complete, statbuf);
+        }
+    }
+    return r_fstatat;
+}
--- a/src/lib/evil/unposix/sys/stat.h
+++ b/src/lib/evil/unposix/sys/stat.h
@@ -10,10 +10,19 @@
 // loading as functions: chmod and umask (for umask, it is necessary to include: errno.h)
 //  fstat ja incluido em <sys/stat.h>
 #include <corecrt_io.h>
+// loading as functions: chmod and umask (for umask, it is necessary to include: errno.h)
+// from corecrt_io import chmod and umask
+// fstat ja incluido em <sys/stat.h>
 
+typedef  int  mode_t ; 
 #ifndef stat64
 # define stat64 _stat64
 #endif
+
+
+int fstatat(int fd, const char *restrict path, struct stat *restrict buf, int flag);
+
+int fstatat64(int fd, const char *restrict path, struct stat *restrict buf, int flag);
 
 // Missing definitions:
 // Note: some pieces of code were based on LibreSSL-Portable's compat lib and
@@ -33,27 +42,12 @@
 # define S_IXOTH  0                           /* Execute others */
 #endif
 
+#define	S_ISFIFO(mode) (((mode) & S_IFMT) == S_IFIFO)
+#define	S_ISDIR(mode)  (((mode) & S_IFMT) == S_IFDIR)
+#define	S_ISREG(mode)  (((mode) & S_IFMT) == S_IFREG)
+#define	S_ISLNK(mode)  (((mode) & S_IFMT) == S_IFLNK)
+#define	S_ISSOCK(mode) (((mode) & S_IFMT) == S_IFSOCK)
+#define	S_ISCHR(mode)  (((mode) & S_IFMT) == S_IFCHR)
+#define	S_ISBLK(mode)  (((mode) & S_IFMT) == S_IFBLK)
+
 #endif
-
-
-
-
-
-
-
-/*
-
-// ------------------------------------------------------------------------
-
-// Sinalizadores de tipo de arquivo para d_type  // ver a questão de st_mode
-
-#define DT_UNKNOWN  0
-#define DT_REG      S_IFREG
-#define DT_DIR      S_IFDIR
-#define DT_FIFO     S_IFIFO
-#define DT_SOCK     S_IFSOCK  //ver 
-#define DT_CHR      S_IFCHR
-#define DT_BLK      S_IFBLK   //ver
-#define DT_LNK      S_IFLNK   //ver 
-
-*/
--- a/src/lib/evil/unposix/sys/stat.h
+++ b/src/lib/evil/unposix/sys/stat.h
@@ -2,9 +2,18 @@
 #define UNPOSIX_SYS_STAT_H
 
 #include <evil_windows.h>
-#include <../ucrt/sys/types.h>
+#include <sys/types.h>
 #include_next <sys/stat.h>
 #include <corecrt.h>
+#include <evil_macro_wrapper.h>
+#include <errno.h>
+// loading as functions: chmod and umask (for umask, it is necessary to include: errno.h)
+//  fstat ja incluido em <sys/stat.h>
+#include <corecrt_io.h>
+
+#ifndef stat64
+# define stat64 _stat64
+#endif
 
 // Missing definitions:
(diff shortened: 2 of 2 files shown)


Coquinho:
how exactly can we test it?

src/lib/evil/stat.c
```diff
@@ -0,0 +1,130 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+//#include <iostream>
+#include <Windows.h>
```
joaoantoniocardoso, line 6: try to use `evil_windows.h` instead, that has the proper WIN_LEAN_AND_MEAN guards to deal with winsock/winsock2 problem.

src/lib/evil/evil_stat.h
```diff
@@ -1,6 +1,33 @@
...
+// fstat ja incluido em <sys/stat.h>
+
+typedef  int  mode_t ; 
+#ifndef stat64
+# define stat64 _stat64
+#endif
+
+int fstatat(int fd, const char *restrict path, struct stat *restrict buf, int flag);
+
+int fstatat64(int fd, const char *restrict path, struct stat *restrict buf, int flag);
```
joaoantoniocardoso, line 29: I think it should be `EAPI int fstatat...`

src/lib/evil/evil_stat.h
```diff
@@ -1,6 +1,33 @@
 #ifndef __EVIL_STAT_H__
 #define __EVIL_STAT_H__
 
+#include <evil_windows.h>
+#include <../ucrt/sys/types.h>
```
joaoantoniocardoso, line 5: Unnecessary here since the removal of `unposix/sys/types.h`.

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,126 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "evil_private.h"
+# include <sys/types.h>
+# include <sys/stat.h>
```
joaoantoniocardoso, line 6: Could you fix this indentation?

src/lib/evil/evil_stat.h
```diff
@@ -1,6 +1,33 @@
...
+#include <sys/types.h>
+#include_next <sys/stat.h>
+#include <corecrt.h>
+#include <evil_macro_wrapper.h>
+#include <errno.h>
+#include <evil_api.h>
+
+// loading as functions: chmod and umask (for umask, it is necessary to include: errno.h)
+//  fstat ja incluido em <sys/stat.h>
```
joaoantoniocardoso, line 14: Could you clean this comments?

src/lib/evil/evil_stat.h
```diff
@@ -1,6 +1,33 @@
...
+
+#include <corecrt_io.h>
+
+// loading as functions: chmod and umask (for umask, it is necessary to include: errno.h)
+// from corecrt_io import chmod and umask
+// fstat ja incluido em <sys/stat.h>
```
joaoantoniocardoso, line 20: Could you clean this comments?

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,126 @@
...
+#include <string.h>
+#include "evil_private.h"
+#include <sys/types.h>
+#include <sys/stat.h>
+#include "evil_stat.h"
+
+#define	AT_SYMLINK_NOFOLLOW	0x01
+
+
+int
```
joaoantoniocardoso, line 12:
```suggestion
EAPI int 
```

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,126 @@
...
+          }
+        else
+          { 
+             r_fstatat = stat(path_complete, statbuf);
+          }
+        return r_fstatat;
+     }
+}
+
+int
```
joaoantoniocardoso, line 70:
```suggestion
EAPI int 
```

src/lib/evil/evil_stat.h
```diff
@@ -1,9 +1,29 @@
 #ifndef __EVIL_STAT_H__
 #define __EVIL_STAT_H__
 
+#include <evil_windows.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <evil_macro_wrapper.h>
+#include <evil_api.h>
+#include <corecrt_io.h>
+#include <corecrt.h>
```
Coquinho, line 10: Shouldn't those libs be inside of `_MSC_VER` guards?

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,126 @@
...
+#include "evil_stat.h"
+
+#define	AT_SYMLINK_NOFOLLOW	0x01
+
+EAPI int 
+fstatat(int dirfd, const char *pathname, struct stat *statbuf, int flags)
+{
+   int r_fstatat;
+
+   if (pathname[2] == '\\')
```
walac, line 17: What if I pass the path `ab\`?
Moreover, we should validate the `pathname` length before accessing. if `pathname` contains `"a"`, we hit a buffer overrun.

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,126 @@
...
+        do
+        {
+           pathbuf_size += MAX_PATH;
+           pathbuf = realloc(pathbuf, pathbuf_size * sizeof(char));
+           copied = GetModuleFileName(NULL, pathbuf, pathbuf_size);
+        } while (copied >= pathbuf_size);
+       
+        pathbuf_size = copied;
+        pathbuf = realloc(pathbuf, pathbuf_size * sizeof(char)); 
+        copied = GetModuleFileName(NULL, pathbuf, pathbuf_size);
```
walac, line 39: Why another call after the loop break?

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,126 @@
...
+
+        int size_str;
+        for (size_str = strlen(pathbuf); size_str >= 0; size_str -- )
```
walac, line 42: Should be `strlen(pathbuf)-1`

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,126 @@
...
+         {
+            if (pathbuf[size_str] == '\\')
+              {
+                 pathbuf[size_str + 1] = 0;
+                 break;
+              }
+         }
```
walac, line 49: What if the path separator is `/`? Furthermore, this loop can be replace by [_splitpath](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/splitpath-wsplitpath?redirectedfrom=MSDN&view=vs-2019)

src/lib/evil/evil_stat.c
walac: And what if the path separator is `/`?

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,126 @@
...
+
+        size_str = strlen(pathbuf) + strlen(pathname);
+
+        char *path_complete = malloc(sizeof(char) * size_str);
+
+        path_complete =  strcat(pathbuf, pathname);
```
walac, line 55: This code makes no sense for me. And it is a potential buffer overrun.

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,126 @@
...
+          {
+             r_fstatat = stat64(path_complete, statbuf);
+          }
+        else
+          { 
+             r_fstatat = stat64(path_complete, statbuf);
+          }
+        return r_fstatat;
+     }
+}
```
walac, line 126: What's the difference in the implementation of `fstatat` and `fstatat64`?

src/lib/evil/evil_stat.c
ricardocvel: if the separator is: "/" or the path is "ab \", simply skip this step, and skip to the next step.

src/lib/evil/evil_stat.c
ricardocvel: okay, I'll fix

src/lib/evil/evil_stat.c
ricardocvel: according to documentation:[http://www.qnx.com/developers/docs/7.0.0/index.html#com.qnx.doc.neutrino.lib_ref/topic/f/fstatat.html](url)
> "The fstatat() and fstatat64() functions obtain information about the file or directory referenced in path. This information  is placed in the structure located at the address indicated by buf. The fstatat64() function is a large-file support version of fstatat()."_

src/lib/evil/evil_stat.c
ricardocvel: so, I based myself on Windows, to put the "/" separator, as for the loop, I will study: _splitpath.

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,126 @@
...
+#include <string.h>
+#include "evil_private.h"
+#include <evil_api.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include "evil_stat.h"
+
+#define	AT_SYMLINK_NOFOLLOW	0x01
+
+EAPI int
```
joaoantoniocardoso, line 12: now it is EVIL_API:
```suggestion
EVIL_API int 
```

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,126 @@
...
+          }
+        else
+          { 
+             r_fstatat = stat(path_complete, statbuf);
+          }
+        return r_fstatat;
+     }
+}
+
+EAPI int
```
joaoantoniocardoso, line 70: now it is EVIL_API:
```suggestion
EVIL_API int 
```

src/lib/evil/evil_stat.c
walac: If the path starts with `C:/`, shouldn't it enter the `if`? And if the path is `ab\`, it will enter the `if`, but if it is `abc\`, it won't. Is this the expected behavior?

src/lib/evil/evil_stat.h
```diff
@@ -1,9 +1,29 @@
...
+
+
+typedef  int  mode_t ; 
+#ifndef stat64
+# define stat64 _stat64
+#endif
+
+EAPI int fstatat(int fd, const char *restrict path, struct stat *restrict buf, int flag);
+
+EAPI int fstatat64(int fd, const char *restrict path, struct stat *restrict buf, int flag);
```
joaoantoniocardoso, line 21: now it is EVIL_API:
```suggestion
EVIL_API int fstatat(int fd, const char *restrict path, struct stat *restrict buf, int flag);

EVIL_API int fstatat64(int fd, const char *restrict path, struct stat *restrict buf, int flag);
```

src/lib/evil/evil_stat.c
walac: I see, but in the Windows implementation they are basically the function, aren't they? If so, why duplicate the code?

src/lib/evil/evil_stat.c
ricardocvel: the idea is that if I want to know, for example, if determined, the name is a file or a directory, and this is not the same as what is being executed.

src/lib/evil/evil_stat.h
ricardocvel: ok, i will correct

src/lib/evil/evil_stat.c
ricardocvel: ok, i will correct

src/lib/evil/evil_stat.c
ricardocvel: ok, i will correct

src/lib/evil/evil_stat.c
ricardocvel: If the path has a "/" in the third character (windows), I understand that the user gave me the full path; in this case, it enters the if.
caso contrário, você entenderá quem é o usuário ou o software que está passando um nome de arquivo ou diretorio, ai não entra no if.
if the user passes paths, such as: "ab / img.jpg" or "abc / img.jpg", it will return -1 entering or not in the if, because it is not a directory name, it is not a file name and not a full file path

src/lib/evil/evil_stat.h
```diff
@@ -19,4 +40,12 @@
 # define S_IXOTH  0                           /* Execute others */
 #endif
 
+#define	S_ISFIFO(mode) (((mode) & S_IFMT) == S_IFIFO)
+#define	S_ISDIR(mode)  (((mode) & S_IFMT) == S_IFDIR)
+#define	S_ISREG(mode)  (((mode) & S_IFMT) == S_IFREG)
+#define	S_ISLNK(mode)  (((mode) & S_IFMT) == S_IFLNK)
+#define	S_ISSOCK(mode) (((mode) & S_IFMT) == S_IFSOCK)
+#define	S_ISCHR(mode)  (((mode) & S_IFMT) == S_IFCHR)
+#define	S_ISBLK(mode)  (((mode) & S_IFMT) == S_IFBLK)
```
Coquinho, line 49: This probably should be inside `_MSC_VER` as cygwin already have those.

src/lib/evil/evil_stat.c
walac: "ab/" is a relative directory name. Furthermore, what if the path is "\Windows"? It is an absolute path, will the function work?

src/lib/evil/evil_stat.c
walac: Is it necessary? If I understood the code, it adds nothing.

src/lib/evil/evil_stat.c
walac: It leaks memory, as the `path_complete` pointer is reassigned, and it shouldn't even compile, as `pathbuf` is `const char *`.

src/lib/evil/evil_stat.c
ricardocvel: In this case, the function returns -1, because a name was passed incorrectly

walac: Why is it incorrect?

ricardocvel: In fact, I expressed myself poorly

I will quote an example: I want to know the size of a given file and, for that, I will use fstatat ():

1. </test/test.txt> in this case, if the file exists and the relative directory exists, it will work correctly.

2. </test/> in this case, if the relative directory exists, it will work correctly and you can get the size of the directory.

3. </test/> in this case, if the relative directory does not exist, the function returns -1.

In this first part of the function, I did just to identify absolute paths.

walac: > In this first part of the function, I did just to identify absolute paths.

The directory `ab\` is a relative path, but the way the code is, it will identify it as an absolute path.

src/lib/evil/evil_stat.c
ricardocvel: I did a test and removed this repeated code.

src/lib/evil/evil_stat.c
ricardocvel: I solved the problem, according to your guidance.
I did a test and it worked as expected

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,62 @@
...
+       
+        int size_str;
+        for (size_str = strlen(pathbuf) -1; size_str >= 0; size_str --)
+         {
+            if (pathbuf[size_str] == '\\')
+              {
+                 pathbuf[size_str + 1] = 0;
+                 break;
+              }
+         }
```
walac, line 45: What about using `_splitpath` instead of a manual for loop?

src/lib/evil/evil_stat.c
```diff
@@ -0,0 +1,62 @@
...
+        strcat(path_complete, pathname);
+
+        if (flags == AT_SYMLINK_NOFOLLOW)
+          {
+             r_fstatat = stat(path_complete, statbuf);
+          }
+        else
+          { 
+             r_fstatat = stat(path_complete, statbuf);
+          }
```
walac, line 59: Why the `if`?

PR #62, opened 2020-05-18
Remove every debugging fprintf/fflush

Coquinho:
It's just a clean-up.


Coquinho:
Maybe we shouldn't apply this PR now, but it's here for when needed.

JPTIZ:
> Maybe we shouldn't apply this PR now, but it's here for when needed.

Why not, exactly?

PR #64, opened 2020-05-20
Change windows expected error message at an ENOMEM error

Coquinho:
Previous it expected "Cannot allocate memory" but windows yields
"Not enough space").

Test Plan:
----------
- The next error will not be related to this;

### DIFF
--- a/src/tests/eina/eina_test_error.c
+++ b/src/tests/eina/eina_test_error.c
@@ -220,8 +220,13 @@ EFL_START_TEST(eina_error_test_failures)
                     EINA_FALSE);
    fail_unless(ctx.did);
 
+#ifdef _MSC_VER
+   ck_assert_str_eq(eina_error_msg_get(ENOMEM),
+                    "Not enough space");
+#else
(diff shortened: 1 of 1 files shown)


src/tests/eina/eina_test_error.c
```diff
@@ -220,8 +220,13 @@ EFL_START_TEST(eina_error_test_failures)
                     EINA_FALSE);
    fail_unless(ctx.did);
 
+#ifdef _MSC_VER
```
walac, line 223: Maybe we should use a definition tied to the platform and not the compiler. The trick part is finding one that's defined for native Windows but not on cygwin.

Coquinho: I only know about `_WIN32` and `_MSC_VER`, from those `_MSC_VER` seams better. @felipealmeida do you know any other definition that we can use here that is better suited?

walac: Let's merge this as-is for now

PR #65, opened 2020-05-20
Eina: Resolve bad comparison while using windows strerror_s

Coquinho:
`strerror_s` is the windows alternative of `strerror_r` used by EFL.

`strerror_s` never return the error code with the message as
`strerror_r` does, because of that, while comparing the first 14
characters of `Unknown error ` to the message from unknown code 4096
(`Unknown error`) they were accusing being different - in UNIX this
works because the message returned is `Unknown error 4096`.

So comparing only the first 13 characters does the trick.

Depends on #64

Test Plan:
----------
- Apply #64;
- The test should pass;

### DIFF
--- a/src/lib/eina/eina_error.c
+++ b/src/lib/eina/eina_error.c
@@ -350,7 +350,7 @@ eina_error_msg_get(Eina_Error error)
                EINA_SAFETY_ERROR("strerror_r() failed");
              else
                {
-                  if (strncmp(str, unknown_prefix, sizeof(unknown_prefix) -1) == 0)
+                  if (strncmp(str, unknown_prefix, sizeof(unknown_prefix) - 2) == 0)
                     msg = NULL;
                   else
                     {


src/lib/eina/eina_error.c
```diff
@@ -350,7 +350,7 @@ eina_error_msg_get(Eina_Error error)
                EINA_SAFETY_ERROR("strerror_r() failed");
              else
                {
-                  if (strncmp(str, unknown_prefix, sizeof(unknown_prefix) -1) == 0)
+                  if (strncmp(str, unknown_prefix, sizeof(unknown_prefix) - 2) == 0)
```
JPTIZ, line 353: For me it makes more sense to actually change the expected error message (since `"Unknown error"` is more a prefix than `"Unknown error "`).

JPTIZ: Also, I would proposa an `is_prefix` function, so you can compare if `unknown_prefix` is a prefix of `str` (i.e. if `"Unknown error"` is prefix of `"Unknown error 1234"`).

felipealmeida: why would you need a function? Isn't this in a test?

JPTIZ: This is in `src/lib/eina/eina_error.c`, not in `src/tests`. Besides, seeing `is_prefix` instead of an inline for-loop seems more reasonable/readable for me.

felipealmeida: We need to look into the goal of a diff. Doing more than what the commit is meant to do is not good. If we want a function that does that, that could be another commit. One that is not high priority for us.

felipealmeida: A good commit does the least amount of modifications needed to accomplish its goal.

PR #69, opened 2020-05-20
Adapt test values converted from [u]int64 to [u]long to avoid overflows 

cauebs:
This and #66 fix the `eina_test_value` test case on Windows.


src/tests/eina/eina_test_value.c
```diff
@@ -159,6 +171,16 @@ EFL_START_TEST(eina_value_test_simple)
    fail_unless(eina_value_ulong_convert(value, &ul));
    fail_unless(ul == (unsigned long)0x1122334455667788);
    fail_unless(u64 == 0x1122334455667788);
+#else
+   fail_unless(eina_value_set(value, 0x12345678));
```
walac, line 175: What about if you add the `ULL` suffix to the constants?

PR #70, opened 2020-05-21
evil_locale.c: Fix infinite recursion call to evil_setlocale

walac:
In the Windows platform, setlocale is defined as an alias to
evil_setlocale, causing an infinite recursive call when category is
different from LC_MESSAGE.

We undef setlocale in the implementation file, which will fall back to
the c runtime function.

### DIFF
--- a/src/lib/evil/evil_locale.c
+++ b/src/lib/evil/evil_locale.c
@@ -17,6 +17,8 @@
  */
 static char _evil_locale_buf[18];
 
+#undef setlocale
+
 char *evil_setlocale(int category, const char *locale)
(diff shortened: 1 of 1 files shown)


felipealmeida:
@vtorri Hmm, can it be evil_private.h is being used incorrectly here, then? I think we're using evil_private.h as include header for all evil includes, instead of being used as an internal header for implementation files of evil.

vtorri:
@felipealmeida again, how do you use setlocale ? in an example outside the tree ? or there is a part of the efl which shows the infinite loop ?

PR #71, opened 2020-05-21
Eina: Missing brackets at eina_vpath()

joaoantoniocardoso:
Due to this missing brackets, the function returns 0 when it in fact executed the 'if' with success.
This is the cause behind `eina_test_vpath` test failing on windows.


Coquinho:
By looking at the code it makes sense, I will try to test this later.

PR #72, opened 2020-05-21
windows: Prevent running the thread before finishing its creation

Coquinho:
CreateThread was running the new thread before the creator thread received
its pointer, causing `fail_if(!eina_thread_equal(t, thread));` to fail
because the variable `thread` wasn't a pointer to the new thread.

Test Plan:
----------
- The next error will not be related to this;
- This test should run until L112 of `_eina_test_lock_thread` at the
  main thread, (here the main thread stops first so I don't know exactly
  the new thread stops yet).

### DIFF
--- a/src/lib/eina/eina_inline_thread_win32.x
+++ b/src/lib/eina/eina_inline_thread_win32.x
@@ -29,13 +29,6 @@ typedef struct _Eina_Win32_Thread_Func
    void *(*func)(void *data);
 } Eina_Win32_Thread_Func;
 
-typedef struct _Eina_Win32_Thread_Attr
-{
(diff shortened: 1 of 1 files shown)


JPTIZ:
`Eina_Thread` is a typedef to `HANDLE` in Windows, is it really a problem? Specially since a thread is created using `CreateThread` (which returns a `HANDLE`) and then stored in a `Eina_Thread` (which is a `HANDLE`):

```c
    *t = CreateThread(thread_attr.lpThreadAttributes, thread_attr.dwStackSize, &_eina_thread_func, &thread_func, thread_attr.dwCreationFlags, threadID);
```

PR #77, opened 2020-05-25
[temporary] Use a custom version of libcheck

walac:
While we don't merge the windows lock fix in the main libcheck, let's
use our fork to keep going with efl port.


JPTIZ:
Tests are still building. Gonna take a deeper look if the thread issue I had (concurrent access to fwrite in libcheck) is solved.

PR #79, opened 2020-05-25
Fix eina_test_file_thread

walac:
This has two fixes: a "use after free" bug fix in the Win32 implementation of eina_thread; and the valid test file for `eina_test_file_thread` test case. See individual commits for details.


JPTIZ:
Ok, I assure file tests don't crash anymore while runnning threads.

Now just gonna have to see why `eina_test_file_thread` doesn't pass.

PR #80, opened 2020-05-25
Windows: get dependencies using NuGet

joaoantoniocardoso:
This PR adds the option to get dependencies for windows using [NuGet](https://www.nuget.org).


native-file-windows.txt
```diff
@@ -11,4 +11,5 @@ endian = 'little'
 [properties]
 c_args = ['-D_WIN32_WINNT=0x0601', '-march=x86-64']
 cpp_args = ['-D_WIN32_WINNT=0x0601', '-march=x86-64']
+nuget-download = true
```
joaoantoniocardoso, line 14:
```suggestion
nuget-download = false
```
At this point we really don't need to enable this, and we could merge without drawbacks.

PR #81, opened 2020-05-26
eina_inline_lock_win32.x: Finish implementation of lock primitives

walac:
The only remaining is spinlock, which is currently implemented with a
mutex/CRITICAL_SECTION. This is will be fixed in a follow up commit.

### DIFF
--- a/src/lib/eina/eina_inline_lock_win32.x
+++ b/src/lib/eina/eina_inline_lock_win32.x
@@ -23,7 +23,7 @@
 # define WIN32_LEAN_AND_MEAN
 #endif
 
-#include <Windows.h>
+#include <windows.h>
 #include <synchapi.h>
 #undef WIN32_LEAN_AND_MEAN
 
@@ -66,7 +66,7 @@ typedef struct _Eina_Lock Eina_Lock;
 typedef struct _Eina_RWLock Eina_RWLock;
 typedef struct _Eina_Condition Eina_Condition;
 typedef struct _Eina_Barrier Eina_Barrier;
-typedef LPDWORD Eina_TLS;
+typedef DWORD Eina_TLS;
 typedef HANDLE Eina_Semaphore;
 
 #ifdef EINA_HAVE_WIN32_SPINLOCK
@@ -81,7 +81,7 @@ EINA_API void eina_lock_debug(const Eina_Lock *mutex);
 typedef HANDLE _Eina_Thread;
 typedef CRITICAL_SECTION _Eina_Mutex_t;
 typedef CONDITION_VARIABLE _Eina_Condition_t;
-typedef PSRWLOCK _Eina_RWLock_t;
+typedef SRWLOCK _Eina_RWLock_t;
 
 struct _Eina_Lock
 {
@@ -100,9 +100,6 @@ struct _Eina_Condition
 {
    Eina_Lock         *lock;      /**< The lock for this condition */
    _Eina_Condition_t condition; /**< The condition variable */
-#if defined(__clockid_t_defined)
-   clockid_t         clkid;     /**< The attached clock for timedwait */
-#endif
 };
 
 enum _Eina_RWLock_Mode {
@@ -114,9 +111,6 @@ enum _Eina_RWLock_Mode {
 struct _Eina_RWLock
 {
    _Eina_RWLock_t mutex; /**< The mutex that handles the locking */
-#ifdef EINA_HAVE_DEBUG_THREADS
-   _Eina_Thread   lock_thread_wid; /**< The ID of the thread that currently has the lock */
-#endif
    enum _Eina_RWLock_Mode mode; /**< The mode the SRWLock is (Shared, Exclusive or None)*/
 };
 /** @} privatesection */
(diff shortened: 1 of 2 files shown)


src/lib/eina/eina_inline_lock_win32.x
```diff
@@ -114,9 +111,6 @@ enum _Eina_RWLock_Mode {
 struct _Eina_RWLock
 {
    _Eina_RWLock_t mutex; /**< The mutex that handles the locking */
-#ifdef EINA_HAVE_DEBUG_THREADS
-   _Eina_Thread   lock_thread_wid; /**< The ID of the thread that currently has the lock */
-#endif
```
Coquinho, line 119: Why removing the debug info?

walac: It wasn't used anyway

src/lib/eina/eina_inline_lock_win32.x
```diff
@@ -209,31 +185,25 @@ _eina_lock_take_try(Eina_Lock *mutex)
...
-   if (eina_lock_take_try(mutex) == EINA_LOCK_SUCCEED) return EINA_LOCK_SUCCEED;
-
    if (_eina_threads_debug >= 100)
      {
         struct timeval t0, t1;
         int dt;
 
         gettimeofday(&t0, NULL);
-        ok = EnterCriticalSection(&(mutex->mutex));
+        EnterCriticalSection(mutex->mutex);
```
Coquinho, line 206: `EnterCriticalSection` receives `LPCRITICAL_SECTION` as arg, so shouldn't it be like:
```suggestion
        EnterCriticalSection(&(mutex->mutex));
```

src/lib/eina/eina_inline_lock_win32.x
```diff
@@ -244,23 +214,11 @@ _eina_lock_take(Eina_Lock *mutex)
...
-   if (ok == ERROR_SUCCESS) ret = EINA_LOCK_SUCCEED;
-   else if (ok == ERROR_POSSIBLE_DEADLOCK)
-     {
-        eina_lock_debug(mutex);
-        ret = EINA_LOCK_DEADLOCK;
-#ifdef EINA_HAVE_DEBUG_THREADS
-        if (_eina_threads_debug) abort();
-#endif
-     }
-   else if (ok != ERROR_TIMEOUT) EINA_LOCK_ABORT_DEBUG((int)ok, lock, mutex);
```
Coquinho, line 262: Why removing the debug stuff?

src/lib/eina/eina_inline_lock_win32.x
```diff
@@ -524,90 +441,35 @@ _eina_tls_get(Eina_TLS key)
...
 _eina_tls_set(Eina_TLS key, const void *data)
 {
-   int ok = TlsSetValue(key, (void *) data);
-   DWORD err = GetLastError();
-   if (ok != 0 && err == ERROR_SUCCESS) return EINA_TRUE;
-   else return EINA_TRUE;
+   return TlsSetValue(key, (void *) data) ? EINA_TRUE : EINA_FALSE;
 }
 
-#ifdef EINA_HAVE_WIN32_BARRIER
```
Coquinho, line 533: If we are removing this, we can remove its definition at `src/lib/eina/meson.build`

src/lib/eina/eina_inline_lock_win32.x
```diff
@@ -524,90 +441,35 @@ _eina_tls_get(Eina_TLS key)
...
-   barrier->needed = needed;
-   barrier->called = 0;
-   if (eina_lock_new(&(barrier->cond_lock)))
-     {
-        if (eina_condition_new(&(barrier->cond), &(barrier->cond_lock)))
-          return EINA_TRUE;
-     }
-   return EINA_FALSE;
-#endif
+   return InitializeSynchronizationBarrier(&barrier->barrier, (LONG) needed, 10)
```
Coquinho, line 455: Why 10 spins? (Just asking, not a suggestion/asking a change)

src/lib/eina/eina_inline_lock_win32.x
walac: Good catch, thanks!

src/lib/eina/eina_inline_lock_win32.x
```diff
@@ -524,90 +441,35 @@ _eina_tls_get(Eina_TLS key)
...
-   DeleteSynchronizationBarrier(&(barrier->barrier));
-   DWORD ok = GetLastError();
-   if (ok != ERROR_SUCCESS) EINA_LOCK_ABORT_DEBUG(ok, barrier_destroy, barrier);
-#else
-   eina_condition_free(&(barrier->cond));
-   eina_lock_free(&(barrier->cond_lock));
-   barrier->needed = 0;
-   barrier->called = 0;
-#endif
+   if (!DeleteSynchronizationBarrier(&barrier->barrier))
```
Coquinho, line 462: [Docs](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-deletesynchronizationbarrier) says that this function always returns `TRUE`, so will it ever enter at this `if`?

src/lib/eina/eina_inline_lock_win32.x
```diff
@@ -524,90 +441,35 @@ _eina_tls_get(Eina_TLS key)
...
+     {
+        EINA_LOCK_ABORT_DEBUG(GetLastError(), barrier_destroy, barrier);
```
Coquinho, line 464: Why removing so many debugging info, but sticking with this one?

src/lib/eina/eina_inline_lock_win32.x
walac: The code no longer takes this code path

src/lib/eina/eina_inline_lock_win32.x
walac: You are right!

src/lib/eina/eina_inline_lock_win32.x
walac: Arbitrary number. It is small enough to not generate contention, but in cases where it is common to be free, it avoids a kernel context switch.

src/lib/eina/eina_inline_lock_win32.x
walac: In future implementations, it may not be the case.

src/lib/eina/eina_inline_lock_win32.x
walac: The ones I removed are code paths no longer valid, but this is.

src/lib/eina/eina_inline_lock_win32.x
Coquinho: Ohh, right, right.

PR #82, opened 2020-05-26
Working on thread join

Coquinho:


Coquinho:
It still has some debugging stuff, I will clean it after lunch.

walac:
Did you update your branch? I see some updates not showing in your PR

src/lib/eina/eina_thread_win32.h
```diff
@@ -37,10 +37,28 @@
...
+typedef struct _Eina_Thread
+{
+   HANDLE handle;
+   struct _Eina_ThreadData *data;
+} Eina_Thread;
 
-#endif 
\ No newline at end of file
+EINA_API DWORD eina_thread_self_id(void);
+EINA_API DWORD eina_thread_id(Eina_Thread t);
```
JPTIZ, line 62: I like the solution, altough I'm worried about having different APIs for Linux and Windows.

I'm talking to the community in `#edevelop` (IRC) to see it further if it would really be a problem.

walac: Hrm, IMHO, there is a design issue here, `eina_evlog` relies on `eina_thread_self` returning an integral type. In Posix systems, it returns a `pthread_t`, which according to Posix [is not guaranteed to be an integral type](https://www.man7.org/linux/man-pages/man3/pthread_self.3.html). My solution would be to introduce the `eina_thread_id` function in the official API and use it when the thread identifier needs to be an integral type, but I would like to get this merged and fixes that in a follow-up PR.

JPTIZ:
As we talked in private, I'll approve this PR so join is done, and later a new PR unifying the API should be done.

PR #88, opened 2020-05-28
Build Eolian

Coquinho:
This PR aims to build `Eolian` itself, but not its unity tests.

This PR changes `EAPI` to `EOLIAN_API` to resolve `Eolian` linking
issues.

It uses `-DEOLIAN_BUILD`, seted at `src/lib/eolian/meson.build`, to
correctly set `EOLIAN_API` as `__declspec(dllexport)` or
`__declspec(dllimport)` at `src/lib/eolian/Eolian.h`.

For the sake of merging it to `native-windows`, this PR does not enable
`Eolian` building or its testes.

If you want to test it, you can reenable `Eolian` build, and mark its tests 
as `false` at `/meson.build`

### DIFF
--- a/src/lib/eolian/Eolian.h
+++ b/src/lib/eolian/Eolian.h
@@ -1,29 +1,25 @@
 #ifndef EOLIAN_H
 #define EOLIAN_H
 
-#ifdef EAPI
-# undef EAPI
+#ifdef EOLIAN_API
+# undef EOLIAN_API
 #endif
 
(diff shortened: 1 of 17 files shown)


joaoantoniocardoso:
> If you want to test it, you can reenable Eolian build, and mark its tests as false at /meson.build

It can be seen here:
https://github.com/expertisesolutions/efl/pull/85/checks?check_run_id=716932021#step:22:808

PR #89, opened 2020-05-28
Merge native-windows to EFL's master

JPTIZ:
Currently, there are some changes being made in EFL's master that are related to our work (or impacts it indirectly), specially those related to the buildsystem. Some of the diffs I can point out (besides changes I saw while rebasing):
- https://phab.enlightenment.org/D11853
- https://phab.enlightenment.org/D11851
- https://phab.enlightenment.org/D11850

**What is most important:** These 3 patches above (which already landed) also means that there's no more need to replace `EAPI`, since it actually works - I was just setting `EFL_BUILD` and `DLL_EXPORT` erroneously. This will also reduce how much we're going to touch EFL's codebase (see TL;DR).

Also, @vtorri has been working on some issues on Windows as he finds them (such as the eina_test_file test, which leads to fixes of wrong usage of Windows API in eina_file_map, for example).

# TL;DR what does this PR solve?
- Lets our current branch up-to-date with EFL master until today, which gives us:
  - No need to change any `EAPI` to `<lib>_API` (or their `*_WEAK_*` "version"): they can be left as they are, just focus on functions and missing features;
  - Simplified sending args to packages (we now just have to change them in main meson.build);
  - Some libraries that are not supposed to be compiled on windows are already marked as so.
- Adds a new schedule function (`eina_sched_prio_init`), since EFL master changed `_eina_internal_call` a bit - but only for posix, so we'll need to implement it to Windows (remind me to open an issue);
- Create a list of ignored subprojects (for more convenience when new changes arrive to subprojects table - and to simplify things as a whole);
- Some minor conflicts with current EFL master.

Also, I commented the File test for eina because it's the only one who currently crashes, but I think we can uncomment when #98 is merged.


walac:
I am more in favor of a merge instead of rebasing since `native-windows` is a public branch. Better yet, we could [cherry pick](https://git-scm.com/docs/git-cherry-pick) the relevant commits instead of a full merge.

JPTIZ:
> I am more in favor of a merge instead of rebasing since `native-windows` is a public branch. Better yet, we could [cherry pick](https://git-scm.com/docs/git-cherry-pick) the relevant commits instead of a full merge.

Considering we already are going to have to throw out the commit tree, maybe a merge would make sense. What I'm worried is about when we have to make a patch for phab.enlightenment.org.

walac:
> Considering we already are going to have to throw out the commit tree, maybe a merge would make sense. What I'm worried is about when we have to make a patch for phab.enlightenment.org.

We will have to rebase our commits when this time arrives.

JPTIZ:
Ok, finally managed to put this altogether into one simple merge commit. No more rebasing needed :) It really simplifies workflow this way haha.

Well, ready for evaluation. I changed the description with a TL;DR of important changes here.

walac:
> Ok, finally managed to put this all together into one simple merge commit. No more rebasing needed :) It really simplifies workflow this way haha.

Won't that make things harder when merging our code to upstream later? With the squash, git can't recognize common commits, making the merge to upstream harder :/

PS: there are some merge conflicts :)

Coquinho:
Oh, I will remove my approval and wait for the resolution of the conflict

JPTIZ:
> > Ok, finally managed to put this all together into one simple merge commit. No more rebasing needed :) It really simplifies workflow this way haha.
> 
> Won't that make things harder when merging our code to upstream later? With the squash, git can't recognize common commits, making the merge to upstream harder :/
> 
> PS: there are some merge conflicts :)

In the end all our work will be squashed in one single commit (because that's what people at phab says to do so: 1 patch, 1 commit), so I think there shouldn't be a problem, or should it?

walac:
> In the end all our work will be squashed in one single commit (because that's what people at phab says to do so: 1 patch, 1 commit), so I think there shouldn't be a problem, or should it?

I never tried the approach of squashing upstream. What about you pick a half dozen commits from native-windows and make a test with a local banch?

JPTIZ:
> > In the end all our work will be squashed in one single commit (because that's what people at phab says to do so: 1 patch, 1 commit), so I think there shouldn't be a problem, or should it?
> 
> I never tried the approach of squashing upstream. What about you pick a half dozen commits from native-windows and make a test with a local banch?

Hmm, nice idea actually. Gonna try that.

JPTIZ:
Ok, did it quickly here. It goes pretty fine: conflicts only goes in actual conflicts I inserted when making this merge (such as adding `if not sys_windows` to disable some things for windows here), in very specific places. The changes that are already on master are ok and causes no conflicts.

To reproduce, my steps were (**given that master's HEAD is pointing to the same commit as this branch's base commit**):

```console
(devs/jptiz/native-rebased) $ git rebase -i <hash of the base commit>
*in editor: pick the base commit, squash every other*
*solve two small conflicts*
(devs/jptiz/native-rebased) $ git rebase --continue
(devs/jptiz/native-rebased) $ git rebase master
```

Coquinho:
@JPTIZ please resolve those conflicts.

JPTIZ:
> @JPTIZ please resolve those conflicts.

Conflicts solved. Tests passes.

Issue #33, opened 2020-04-09
Undefine _WIN32_LEAN_AND_MEAN after use


Coquinho:
Because we don't know if `_WIN32_LEAN_AND_MEAN` can scramble something, and for organizational purposes, we should undefine it as soon as the includes that it is used to have happened. This change should happen at every file that defines `_WIN32_LEAN_AND_MEAN`.

So something like:
```
#ifdef _WIN32_
# if !defined(_WIN32_LEAN_AND_MEAN)
#  define _WIN32_LEAN_AND_MEAN
# endif
#endif
<include that uses  WIN API>
...
<last include that uses  WIN API>
#undef _WIN32_LEAN_AND_MEAN
```

PR #91, opened 2020-05-29
Fixing some windows includes with WIN32_LEAN_AND_MEAN

joaoantoniocardoso:
Today some sources are including `<windows.h>` without defining `WIN32_LEAN_AND_MEAN`, which in our case is leading to the [well known problem about winsock vs winsock2](https://stackoverflow.com/a/9168850). This PR fixes all `<Windows.h>`, following the recommendations of #33.

### DIFF
--- a/src/lib/ecore_win32/ecore_win32_clipboard.c
+++ b/src/lib/ecore_win32/ecore_win32_clipboard.c
@@ -2,10 +2,6 @@
 # include <config.h>
 #endif
 
-#define WIN32_LEAN_AND_MEAN
-#include <windows.h>
-#undef WIN32_LEAN_AND_MEAN
-
 #include <evil_private.h> /* utf-8 and utf-16 conversion */
 #include <Eina.h>
 
--- a/src/lib/ecore_win32/ecore_win32_cursor.c
+++ b/src/lib/ecore_win32/ecore_win32_cursor.c
@@ -6,7 +6,7 @@
 # define WIN32_LEAN_AND_MEAN
 #endif
 #include <windows.h>
-#include <windowsx.h>
+#include <winuser.h>
 #undef WIN32_LEAN_AND_MEAN
 
 #include <Eina.h>
--- a/src/lib/ecore_win32/ecore_win32_cursor.c
+++ b/src/lib/ecore_win32/ecore_win32_cursor.c
@@ -2,8 +2,11 @@
 # include <config.h>
 #endif
 
-#define WIN32_LEAN_AND_MEAN
+#ifndef WIN32_LEAN_AND_MEAN
+# define WIN32_LEAN_AND_MEAN
+#endif
 #include <windows.h>
+#include <windowsx.h>
 #undef WIN32_LEAN_AND_MEAN
 
 #include <Eina.h>
--- a/src/lib/eina/eina_log.c
+++ b/src/lib/eina/eina_log.c
@@ -21,10 +21,6 @@
 # include "config.h"
 #endif
 
-#ifdef _WIN32
-# include <evil_private.h> /* vasprintf */
-#endif /* _WIN32 */
-
 #include <string.h>
 #include <stdlib.h>
 #include <unistd.h>
@@ -36,7 +32,6 @@
 #endif
 
 #ifdef _WIN32
-# include <windows.h>
 # include <evil_private.h>
 #endif
 
--- a/src/lib/eina/eina_prefix.c
+++ b/src/lib/eina/eina_prefix.c
@@ -42,8 +42,8 @@
 #endif
 
 #ifdef _WIN32
-# include <direct.h> /* getcwd */
 # include <evil_private.h> /* path_is_absolute realpath dladdr */
+# include <direct.h> /* getcwd */
 #endif
 
 #include "eina_config.h"
--- a/src/lib/elementary/elm_main.c
+++ b/src/lib/elementary/elm_main.c
@@ -8,9 +8,9 @@
 
 #ifdef _WIN32
 # include <evil_private.h> /* dlopen,dlclose,etc */
-# include <dlfcn.h> /* dlopen,dlclose,etc */
 # include <direct.h> /* getcwd */
 #else
+# include <dlfcn.h> /* dlopen,dlclose,etc */
 # include <unistd.h>
 #endif
 
--- a/src/lib/elementary/elm_main.c
+++ b/src/lib/elementary/elm_main.c
@@ -2,17 +2,14 @@
 # include "elementary_config.h"
 #endif
 
-#ifndef _WIN32
-# include <dlfcn.h> /* dlopen,dlclose,etc */
-#endif
-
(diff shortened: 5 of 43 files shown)


src/lib/ecore_win32/ecore_win32_clipboard.c
```diff
@@ -2,10 +2,6 @@
 # include <config.h>
 #endif
 
-#define WIN32_LEAN_AND_MEAN
-#include <windows.h>
-#undef WIN32_LEAN_AND_MEAN
-
```
Coquinho, line 8: Why removing it?

src/lib/ecore_win32/ecore_win32_cursor.c
```diff
@@ -2,8 +2,11 @@
 # include <config.h>
 #endif
 
-#define WIN32_LEAN_AND_MEAN
+#ifndef WIN32_LEAN_AND_MEAN
+# define WIN32_LEAN_AND_MEAN
+#endif
 #include <windows.h>
+#include <windowsx.h>
```
Coquinho, line 9: Why now we need `windowsx.h` here?

src/lib/eina/eina_log.c
```diff
@@ -36,7 +32,6 @@
 #endif
 
 #ifdef _WIN32
-# include <windows.h>
```
Coquinho, line 39: Why removing the explicit `windows.h` include?

src/lib/eina/eina_prefix.c
```diff
@@ -42,8 +42,8 @@
 #endif
 
 #ifdef _WIN32
-# include <direct.h> /* getcwd */
 # include <evil_private.h> /* path_is_absolute realpath dladdr */
+# include <direct.h> /* getcwd */
```
Coquinho, line 46: Are there any problems with the other order?

src/lib/elementary/elm_main.c
```diff
@@ -2,17 +2,14 @@
...
-#endif
-
 #ifdef HAVE_CRT_EXTERNS_H
 # include <crt_externs.h>
 #endif
 
 #ifdef _WIN32
-# include <direct.h> /* getcwd */
 # include <evil_private.h> /* dlopen,dlclose,etc */
+# include <dlfcn.h> /* dlopen,dlclose,etc */
```
Coquinho, line 11: This should go to the `#else` side, as it should be included only when not on windows.

src/lib/evil/evil_windows.h
```diff
@@ -8,11 +8,9 @@
...
 # define WIN32_LEAN_AND_MEAN
 #endif
-
-#include <evil_api.h>
-
 #include <windows.h>
-#include <winsock2.h>
+#undef WIN32_LEAN_AND_MEAN
 
+#include <evil_api.h>
```
Coquinho, line 14: Why putting `evil_api.h` on the place of `winsock2.h`? `evil_api.h` doesn't seam to include `winsock2.h` nor does what `winsock2.h` should do.

walac:
Instead of adding the define in every single include, why not pass it to the compiler inside `meson.build`?

src/lib/ecore_win32/ecore_win32_cursor.c
joaoantoniocardoso: oops, actually it uses `winuser.h`, not `windowsx.h`.
```suggestion
#include <winuser.h>
```

src/lib/eina/eina_log.c
joaoantoniocardoso: > `evil_private.h` includes `<evil_windows.h>`, which uses `WIN32_LEAN_AND_MEAN`.

src/lib/eina/eina_prefix.c
joaoantoniocardoso: I didn't test with this particular case but if `<direct.h>` includes `<windows.h>`, possibly it will  mess with our `WIN32_LEAN_AND_MEAN`, that is used by `<evil_private.h>` include.

src/lib/evil/evil_fcntl.c
```diff
@@ -2,14 +2,14 @@
 # include "config.h"
 #endif /* HAVE_CONFIG_H */
 
-#include <evil_windows.h>
+#include "evil_private.h"
+#include <winsock.h>
```
joaoantoniocardoso, line 6: it should be `winsock2` here.

src/lib/eina/eina_prefix.c
Coquinho: oh, okay

src/lib/evil/evil_windows.h
joaoantoniocardoso: `evil_api.h` was moved below windows.h and I am not sure about `winsock2` but it seems not always needed. Just to remember, `evil_windows.h` is included in `evil_private.h`.

Coquinho: but looking at the diff, it shows that `winsock2.h` was removed.

joaoantoniocardoso: Yes it is removed, do you think it should be always included with `windows.h`?

joaoantoniocardoso:
> Instead of adding the define in every single include, why not pass it to the compiler inside `meson.build`?

I agree with this approach if what #33 stated is not a real problem. Also I think it should be well documented to avoid some confusion/messy flag configs.

src/lib/evil/evil_windows.h
Coquinho: I have no idea about it, but it was included before, so we *expect* it to be necessary.
As this PR just wants to protect `windows.h` of adding unwanted headers, maybe it shouldn't remove this, unless it is causing an error.

src/lib/ecore_win32/ecore_win32_clipboard.c
Coquinho: I just don't know if we should remove explicitly includes, as it makes things a little more clear. But I know it will work this way.

src/lib/evil/evil_windows.h
joaoantoniocardoso: Agreed. Redundancy is not a problem.

src/lib/ecore_win32/ecore_win32_cursor.c
Coquinho: hummm, and why we need `winuser.h`?

Coquinho: I mean, it seems that it wasn't needed before, so why it is needed now?

joaoantoniocardoso: There are several functions from `winuser.h` been used here, maybe `windows.h` includes it already?

src/lib/evil/evil_windows.h
joaoantoniocardoso: My bad. I tested again and in fact `<winsock2.h>` really needs to be removed. I will rebase to keep on the right track.

PR #96, opened 2020-05-29
eina: Implement the eina_thread_self_id API

walac:
eina_self_thread returns the type Eina_Thread, which is not guaranteed
to return a type comparable with the == operator or convertable to a
integral type.

This commits introduces the eina_thread_self_id API, which returns a
thread identifier compatible to an integral type.

### DIFF
--- a/src/lib/eina/eina_inline_thread_posix.x
+++ b/src/lib/eina/eina_inline_thread_posix.x
@@ -18,6 +18,14 @@
 
 #include "eina_thread_posix.h"
 
+#ifdef HAVE_PTHREAD_NP_H
+#include <pthread_np.h>
+#endif
+
+#ifdef __linux__
+#include <sys/types.h>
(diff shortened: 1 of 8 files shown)


Coquinho:
Oh, just so that I read it wrong, I will reevaluate.

walac:
> Another option that I see would be implementing an eina_thread_id as I said early.

I tried that, but the Win32 API is the only one that permits getting the thread ID from the handle. Doing that for Posix would imply changing the `Eina_Thread` definition from an integral to a `struct`, which I am afraid could break a lot of user code out there.

walac:
> we should remove the Eina_Thread that is an arg in many functions here

That could go in a follow up PR.

walac:
> That could go in a follow up PR.

Giving a little more thought on this, I think it makes sense to implement in this PR. Done.

PR #98, opened 2020-06-01
windows: fix eina_file_map_new

walac:
On Windows, the file map offset must be a multiple of the system allocation granularity.

We handle the alignment requirement internally. We round down the given offset to the next
aligned value. We then map the aligned offset plus the offset difference plus the map length.
We adjust the returned pointer value to the original offset and store the original pointer inside
the file map structure for use by the unmap function.

### DIFF
--- a/src/lib/eina/eina_file_win32.c
+++ b/src/lib/eina/eina_file_win32.c
@@ -917,6 +917,10 @@ eina_file_map_new(Eina_File *file, Eina_File_Populate rule,
    if (!map)
      {
         HANDLE fm;
+        SYSTEM_INFO sys_info;
+
+        GetSystemInfo(&sys_info);
+
         map = malloc(sizeof (Eina_File_Map));
         if (!map)
           {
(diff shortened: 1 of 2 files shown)


src/tests/eina/eina_test_file.c
```diff
@@ -29,6 +29,7 @@
 
 #ifdef _WIN32
 # include <evil_private.h> /* mkdir */
+# include <windows.h>
```
joaoantoniocardoso, line 32: `evil_private.h` is indirectly including `windows.h` already. I would suggest to use `WIN32_LEAN_AND_MEAN` if you want to explicitly keep it here.

JPTIZ:
@vtorri was checking this before. I'll ask him to see if these are all necessary changes as well until giving a full approve (besides what @joaoantoniocardoso already said).

Anyway, for me here, it indeed fixes `eina_file_map_new` :)

vtorri:
i don' think that the user must use granularity, instead eina_file_map_new should manage that

i have a code that does that, but the map is not completely correct. I have to fix that

vtorri:
also, what annoys me is that the test is rewritten so that the code works (especially with the *2 of the size). file mapping must work with whatever size the user want.

walac:
> i don' think that the user must use granularity, instead eina_file_map_new should manage that

I don't see how we can handle this internally. [MapViewOfVile](https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffile) will fail if the offset isn't a multiple of the granularity. The solution I see is adjusting the offset and pointer internally and storing inside the `Eina_File` structure to be used when unmapping the view.

vtorri:
in eina_file_map_new : 
 - you compute the largest offset that is multiple of granularity and <= offset : start = (offset/ granulariity) * granularity
 - you returns mapped data and you add to it this value : (offset - start), which is the offset value passed to eina_file_map_new

walac:
> you return mapped data and you add to it this value : (offset - start), which is the offset value passed to eina_file_map_new

That's my idea plus: store the offset/pointer pair information inside the `Eina_File` structure so we can adjust the pointer back to pass to [UnmapViewOfFile](https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-unmapviewoffile). Does that make sense to you?

src/tests/eina/eina_test_file.c
walac: Done.

vtorri:
it totally does makesense

 including windows.h after evil_private.h will have no effect as it is included before in evil_private.h.

walac:
@vtorri I updated the patch with your idea

src/lib/eina/eina_file_win32.c
```diff
@@ -917,24 +917,35 @@ eina_file_map_new(Eina_File *file, Eina_File_Populate rule,
...
              return NULL;
           }
 
+        const DWORD align = ~(sys_info.dwAllocationGranularity - 1);
+        const DWORD aligned_offset = offset & align;
+        const DWORD offset_diff = offset - aligned_offset;
+        const uint64_t total_length = offset + length;
+        const DWORD length_low = total_length & (uint64_t) (DWORD)-1;
+        const DWORD length_high = total_length >> (sizeof(DWORD) * 8);
+
```
Coquinho, line 937: Could you explain a little what is happening here? I couldn't get it.

walac: The typical alignment granularity for Windows is 64kb. That means the `MapViewOfFile` offset must be a multiple of 64kb. To return the proper pointer, we must find the offset that is aligned by 64kb that is equal or less than the passed offset. `align` is the mask to find the aligned offset. `aligned_offset` is the final offset value. `offset_set` is the distance from the `aligned_offset` to the passed `offset`.

src/lib/eina/eina_file_win32.c
```diff
@@ -917,24 +917,35 @@ eina_file_map_new(Eina_File *file, Eina_File_Populate rule,
...
+        GetSystemInfo(&sys_info);
+
         map = malloc(sizeof (Eina_File_Map));
         if (!map)
           {
              eina_lock_release(&file->lock);
              return NULL;
           }
 
+        const DWORD align = ~(sys_info.dwAllocationGranularity - 1);
```
cauebs, line 931: Perhaps it would be best to include some comments explaining the logic behind these operations. I have a vague notion of what's being done and I'm having trouble understanding this. It might seem even more obscure to future readers.

PR #100, opened 2020-06-01
add missing "!" to thread inequality

Coquinho:
[By comparing ](https://github.com/expertisesolutions/efl/compare/devs/felipealmeida/eolian-test...devs/cauebs/eo-build) we can see that the original code was:
```
   if ((self != _efl_object_main_thread) &&	
       (self != klass->construction_thread))	
```
So the desired behavior would be:
```
   if (!eina_thread_equal(self, _efl_object_main_thread) &&
       !eina_thread_equal(self, klass->construction_thread))
```


cauebs:
Nice catch!

PR #101, opened 2020-06-01
windows: Fix path for file sanitization

walac:


src/tests/eina/eina_test_file.c
```diff
@@ -527,7 +527,11 @@ static const struct {
    const char *test;
    const char *result;
 } sanitize[] = {
+#if defined _WIN32
+  { "C:/home/mydir/../myfile", "C:/home/myfile" }
```
joaoantoniocardoso, line 531: Wouldn't be better to use `\\` instead of `/`?

Coquinho: oh, good point, this change makes the `eina_test_file_path` pass through.

Coquinho: @walac do you know why using `/` is working?

walac: Windows supports `/` as a path separator. I am working on more test cases for this.

joaoantoniocardoso: Indeed but I don't see as an option of choice for win32 api as its functions will only return `\\` as path separators.

walac: I am not sure I am following. In the updated patch I added tests for the `\` seperator cases.

joaoantoniocardoso: yeah, the updated is ok to me.

PR #102, opened 2020-06-01
eina_test_file.c: Do not check errno on success

walac:
errno is documented to only contains a valid value for mkstemp if the
function fails.

### DIFF
--- a/src/tests/eina/eina_test_file.c
+++ b/src/tests/eina/eina_test_file.c
@@ -790,7 +790,6 @@ EFL_START_TEST(eina_test_file_mktemp)
         fd = eina_file_mkstemp(patterns[k], &tmpfile);
         ck_assert(fd >= 0);
         ck_assert(!!tmpfile);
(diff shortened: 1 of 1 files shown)


Coquinho:
@walac are you sure you are approaching it correctly? I mean, at windows errno 17 is returned, but at Linux, 0 is returned. Maybe the error here is in another point.

walac:
> are you sure you are approaching it correctly? I mean, at windows errno 17 is returned ("File Exists"), but at Linux 0 is returned as there is no error. Maybe the error isn't here but in another point

An API is free to internally call some system call that fails and set `errno`, and leave it as it is. If the return value of a function says it succeeded, the `errno` value is irrelevant.

Coquinho:
hummm, that makes sense, but which function set errno? The only function that could set it and that is called after resting it is eina_file_mkstemp isn't it?  And it seems that this value (17) is happening inside `eina_file_mkstemp`

PR #103, opened 2020-06-02
Make #89 compilation process achieve the same point of native-windows on Linux

Coquinho:
The errors were related to:
- Lua was disabled: by just re-enabling it could go further;
- `master` removed the `systemd` option from `ecore`;

### DIFF
--- a/src/lib/ecore/meson.build
+++ b/src/lib/ecore/meson.build
@@ -193,10 +193,6 @@ if get_option('g-mainloop') == true
   endif
 endif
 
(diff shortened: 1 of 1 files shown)


Coquinho:
If this PR is approved after #89 have been landed, I will change it to `devs/expertise/native-windows`

Coquinho:
By adding the #89 Linux is no more building.
This PR still make sense thought part of its original change is already on `native-windows`

PR #104, opened 2020-06-03
Linux Build: Make Nugget variable only on windows build

Coquinho:
Nugget variable was being set on every builds, causing an error at linux
build while configuring the build, now it successfully configures
build.


Coquinho:
I'm testing it with Travis-CI at my [fork](https://github.com/Coquinho/efl/runs/735115819), if you want to have a look at it.

PR #106, opened 2020-06-03
Linux Build: pipe function were not available at a linux build

Coquinho:
As pipe functions were defined at `evil` they were not available to any
other library. Now they are defined at `eina`.

You can have a look at tests with Travis-CI at my [fork](https://github.com/Coquinho/efl/runs/735172415).

Depends on #104


Coquinho:
This branch has been rebased as #104 have been merged to `native-windows`

Issue #58, opened 2020-05-08
No `clock_gettime()` implementation available for `_WIN32` at `src/tests/eina/eina_test_lock.c`.


joaoantoniocardoso:
Building in _WIN32, the test source `src/tests/eina/eina_test_lock.c` calls `clock_gettime()` _but there is no implementation available for `_WIN32`.

For linux, glibc implements `clock_gettime()` at `<time.h>` and for `__MACH__` `src/tests/eina/eina_test_lock.c` it self has an implementation using `<mach/clock.h>`.

note: I have no clue about the requirements for time precision.

vtorri:
this should be replaced with high-resolution time stamps with win32

felipealmeida:
Thanks vtorry

vtorri:
do you want some code for that ?

PR #107, opened 2020-06-03
windows: Fix tests cases in eina_test_lock.c

walac:
We implement the `clock_gettime` for Windows and set eina error when the condition variable wait times out.

### DIFF
--- a/src/lib/eina/eina_inline_lock_win32.x
+++ b/src/lib/eina/eina_inline_lock_win32.x
@@ -300,6 +300,10 @@ _eina_condition_wait_ex(Eina_Condition *cond, DWORD timeout)
           {
              EINA_LOCK_ABORT_DEBUG((int) err, cond_wait, cond);
           }
+        else if (ERROR_TIMEOUT == err)
+          {
+             eina_error_set(ETIMEDOUT);
+          }
      }
 
 #ifdef EINA_HAVE_DEBUG_THREADS
--- a/src/tests/eina/eina_test_lock.c
+++ b/src/tests/eina/eina_test_lock.c
@@ -60,7 +60,14 @@ clock_gettime(int mode, struct timespec* ts)
 int
 clock_gettime(int mode, struct timespec* ts)
 {
-// TODO: implement clock_gettime() for _WIN32.
+   FILETIME sys_time;
+   ULARGE_INTEGER li_sys_time;
+
+   GetSystemTimeAsFileTime(&sys_time);
+   li_sys_time.u.LowPart = sys_time.dwLowDateTime;
+   li_sys_time.u.HighPart = sys_time.dwHighDateTime;
+   ts->tv_sec = li_sys_time.QuadPart / 10000000UL;
+   ts->tv_nsec = (li_sys_time.QuadPart % 10000000UL) * 100UL;
    return 0;
 }
 


src/lib/eina/eina_inline_lock_win32.x
```diff
@@ -300,6 +300,10 @@ _eina_condition_wait_ex(Eina_Condition *cond, DWORD timeout)
           {
              EINA_LOCK_ABORT_DEBUG((int) err, cond_wait, cond);
           }
+        else if (ERROR_TIMEOUT == err)
+          {
+             eina_error_set(ETIMEDOUT);
+          }
```
Coquinho, line 306: What needs a time out error?

src/tests/eina/eina_test_lock.c
```diff
@@ -60,7 +60,14 @@ clock_gettime(int mode, struct timespec* ts)
 int
 clock_gettime(int mode, struct timespec* ts)
 {
-// TODO: implement clock_gettime() for _WIN32.
+   FILETIME sys_time;
+   ULARGE_INTEGER li_sys_time;
+
+   GetSystemTimeAsFileTime(&sys_time);
```
Coquinho, line 66: I was looking #58, @vtorri recommended using a [high-resolution timestamp](https://docs.microsoft.com/en-us/windows/win32/sysinfo/acquiring-high-resolution-time-stamps), so shouldn't we use [GetSystemTimePreciseAsFileTime](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsystemtimepreciseasfiletime?redirectedfrom=MSDN) here?

vtorri:
QueryPerformanceFrequency  and QueryPerformanceCounter

Coquinho:
> QueryPerformanceFrequency and QueryPerformanceCounter

@vtorri would it be a problem to use `GetSystemTimePreciseAsFileTime` instead? It seems to do the trick here.

src/tests/eina/eina_test_lock.c
vtorri: @Coquinho note that this function is only available on windows >= 8

vtorri:
i've just commented that this function is available on win >= 8

also, note that we don't want clock_gettime-like function. clock_gettime is just to measure the time spent but some code. The time itself is not used.

src/tests/eina/eina_test_lock.c
walac: I tried to use `QueryPerformanceCounter`. As it uses the number of ticks, the resolution makes the test failure (`49 < 50`). `GetSystemTimeAsFileTime` does the job, as its granularity is of the order of 100 ns.

vtorri: can you show me your code with QPF/QPC, please ?

walac: Here it is
```diff
diff --git a/src/tests/eina/eina_test_lock.c b/src/tests/eina/eina_test_lock.c
index 3b923ee108..053b307c1c 100644
--- a/src/tests/eina/eina_test_lock.c
+++ b/src/tests/eina/eina_test_lock.c
@@ -57,17 +57,20 @@ clock_gettime(int mode, struct timespec* ts)

 #  define CLOCK_REALTIME 0

+#define NS 1000000000
+
 int
 clock_gettime(int mode, struct timespec* ts)
 {
-   FILETIME sys_time;
-   ULARGE_INTEGER li_sys_time;
-
-   GetSystemTimeAsFileTime(&sys_time);
-   li_sys_time.u.LowPart = sys_time.dwLowDateTime;
-   li_sys_time.u.HighPart = sys_time.dwHighDateTime;
-   ts->tv_sec = li_sys_time.QuadPart / 10000000UL;
-   ts->tv_nsec = (li_sys_time.QuadPart % 10000000UL) * 100UL;
+   LARGE_INTEGER sys_time, frequency;
+
+   if (!QueryPerformanceCounter(&sys_time) || !QueryPerformanceFrequency(&frequency))
+       return -1;
+
+   sys_time.QuadPart *= NS;
+   sys_time.QuadPart /= frequency.QuadPart;
+   ts->tv_sec = sys_time.QuadPart / NS;
+   ts->tv_nsec = sys_time.QuadPart - ts->tv_sec * NS;
    return 0;
 }

@@ -253,8 +256,8 @@ EFL_START_TEST(eina_test_rwlock)
    eina_condition_timedwait(&cond, 0.050);
    clock_gettime(CLOCK_REALTIME, &ts2);
    delay = (ts2.tv_sec - ts.tv_sec) * 1000L + (ts2.tv_nsec - ts.tv_nsec) / 1000000L;
-   fail_if(delay < 50);
-   fail_if(delay > 200);
+   ck_assert_msg(delay >= 50, "%ld < 50", delay);
+   ck_assert_msg(delay <= 200, "%ld > 200", delay);
    fail_if(eina_error_get() != ETIMEDOUT);
    fail_if(eina_lock_release(&mtcond) != EINA_LOCK_SUCCEED);
```

vtorri: - compute frequency once
 - QueryPerformanceCounter and QueryPerformanceFrequency can't fail with windows  > xp
 - you can also not use a clock_gettime function on Windows, but something different (we are in a test)
 - so : 
   - compute frequency before
   - function which returns the counter
   - compute the difference when calling this function twice, divide it by frequency

walac: Do you mean like so?
```diff
diff --git a/src/tests/eina/eina_test_lock.c b/src/tests/eina/eina_test_lock.c
index 3b923ee108..dfd3f0808b 100644
--- a/src/tests/eina/eina_test_lock.c
+++ b/src/tests/eina/eina_test_lock.c
@@ -34,6 +34,8 @@

 #  define CLOCK_REALTIME 0

+#include <math.h>
+
 int
 clock_gettime(int mode, struct timespec* ts)
 {
@@ -213,7 +215,6 @@ _eina_test_rwlock_thread(void *data EINA_UNUSED, Eina_Thread t EINA_UNUSED)

 EFL_START_TEST(eina_test_rwlock)
 {
-   struct timespec ts, ts2;
    long delay;


@@ -249,10 +250,22 @@ EFL_START_TEST(eina_test_rwlock)
    fail_if(eina_rwlock_release(&mutex) != EINA_LOCK_SUCCEED);

    fail_if(eina_lock_take(&mtcond) != EINA_LOCK_SUCCEED);
+#ifdef _WIN32
+   LARGE_INTEGER start, end, frequency;
+   QueryPerformanceFrequency(&frequency);
+   QueryPerformanceCounter(&start);
+#else
+   struct timespec ts, ts2;
    clock_gettime(CLOCK_REALTIME, &ts);
+#endif
    eina_condition_timedwait(&cond, 0.050);
+#ifdef _WIN32
+   QueryPerformanceCounter(&end);
+   delay = ceil(((end.QuadPart - start.QuadPart) * 1000.0) / (double) frequency.QuadPart);
+#else
    clock_gettime(CLOCK_REALTIME, &ts2);
    delay = (ts2.tv_sec - ts.tv_sec) * 1000L + (ts2.tv_nsec - ts.tv_nsec) / 1000000L;
+#endif
    fail_if(delay < 50);
    fail_if(delay > 200);
    fail_if(eina_error_get() != ETIMEDOUT);
```

vtorri: yes, is it working, now ?

walac: Yes, it works! But I wonder if it is worth making the code more complex for a test just for the sake of using another API. However, I have no strong feelings here, if you prefer this, I will update the patch.

vtorri: keep the port of clock_gettime(), it's just a test anyway. It would be nice if you can create a patch on phab

PR #108, opened 2020-06-03
Linux Build: Correct a typo

Coquinho:
A typo was yielding a linking error.


Coquinho:
Oh, and here is the [travis log](https://travis-ci.com/github/Coquinho/efl/builds/169480597) (just passed there)

PR #113, opened 2020-06-04
windows: Implement delete callback funcionality for tls

walac:
The implementation makes use of the DLL_THREAD/PROCESS_DETACH to call
delete_cb when the thread exits.

### DIFF
--- a/src/lib/eina/eina_inline_lock_win32.x
+++ b/src/lib/eina/eina_inline_lock_win32.x
@@ -48,6 +48,8 @@ typedef void (*Eina_Lock_Bt_Func) ();
 #include "eina_inlist.h"
 #endif
 
+#include "eina_hash.h"
+
 EINA_API void _eina_lock_debug_abort(int err, const char *fn, const volatile void *ptr);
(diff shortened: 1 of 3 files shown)


walac:
> I don't know if this could have an impact here, but evil has a DLLMain function too.

No, they are different DLLs.

Coquinho:
@waluc how can I test this? Is there any unity test?

Other thing, about naming, at [EFL code conventions](https://www.enlightenment.org/contrib/devs/coding-conventions.md) 
its said to name things as `{module}_{object}_{specializations}_{action}`, so
the best probably would be in ways like:
- `_free_tls_value` -> `_eina_tls_value_free`
- `_tls_hash_foreach_cb` -> `_eina_tls_hash_foreach_cb`
- `eina_win32_dllmain.c` -> `eina_tls_win32.c`

So, what do you guys (@JPTIZ, @joaoantoniocardoso, @walac) think about the naming thing?

Sorry about my other (wrong) suggestion, @walac. :(

PS: for now my approve is more bonded over the tests than the code
conventions, as we may address it latter;

walac:
> how can I test this? Is there any unity test?

`eina_test_lock`

I've updated the code with your latest comments addressed.

PR #109, opened 2020-06-03
[Deps] eo+eolian+efl

joaoantoniocardoso:
Branching from `devs/felipealmeida/native-windows/eolian-api` (no more `EFL_EXTERN` thing needed here).

What is building:
- eolian
  - [x] lib
  - [x] bin
  - [x] tests
- eo
  - [x] lib
  - [x] tests
- efl
  - [x] lib
  - [x] tests: not enabled, _depends on ecore_

PR #115, opened 2020-06-04
[Deps] ecore

joaoantoniocardoso:
After eo+eolian+efl (#109), this PR starts to work with ecore. 

The strategy adopted here was to bypass all errors to compile ecore, and then other dependencies could be started from here before ecore is ready.

The following list represents the errors that needs to be fixed, each one is unrelated and can be done independently:

- [x] Bypassing unneeded(?) header for windows c4d23b0, fixed by 00dc563
- [x] Fix-me: Stub added for ftruncate 49198de
  - Fixed by: b7e1f25
- [ ] Fix-me: By-passing ecore_thread errors 2fe441e
  - Possibly fixed with [D12037](https://phab.enlightenment.org/D12037) 
- [ ] By-passing HAVE_GETTIMEOFDAY checking 6561eb5
- [ ] By-passing gettimeofday undefined errors 95e6776


Coquinho:
Nice, your description is really nice, well done.

Coquinho:
[95e6776](https://github.com/expertisesolutions/efl/commit/95e677698dde978776911cac349e12afc768392b) seems to me as a good solution to your both problems with `gettimeofday`, just need the `EAPI/EVIL_API` thing

joaoantoniocardoso:
> [95e6776](https://github.com/expertisesolutions/efl/commit/95e677698dde978776911cac349e12afc768392b) seems to me as a good solution to your both problems with `gettimeofday`, just need the `EAPI/EVIL_API` thing

Yeah it could be, just remember that also related to gettimeofday there is the error of `HAVE_GETTIMEOFDAY` not being defined (see 
e5cf5f5 ).

I think that the flag `HAVE_GETTIMEOFDAY` could be defined together with the declaration of gettimeofday using `_WIN32` guards, but I remember to face see some other errors when tried this.. That's why I just bypassed this checks. 

On the other hand, If bypassing is a viable option, one should analyze other sources that are using this flag too.

Coquinho:
> > [95e6776](https://github.com/expertisesolutions/efl/commit/95e677698dde978776911cac349e12afc768392b) seems to me as a good solution to your both problems with `gettimeofday`, just need the `EAPI/EVIL_API` thing
> 
> Yeah it could be, just remember that also related to gettimeofday there is the error of `HAVE_GETTIMEOFDAY` not being defined (see [e5cf5f5](https://github.com/expertisesolutions/efl/commit/e5cf5f578c9041f406b7eb704fdda338784dbc72)).
> 
> I think that the flag `HAVE_GETTIMEOFDAY` could be defined together with the declaration of gettimeofday using `_WIN32` guards, but I remember to see some other errors when tried this.. That's why I just bypassed this check.

`HAVE_` definitions are done at `meson.build`, so we could define this at `evil/meson.build`

> On the other hand, If bypassing is a viable option, one should analyze other sources that are using this flag too.

joaoantoniocardoso:
> HAVE_ definitions are done at meson.build, so we could define this at evil/meson.build

Nice, I didn't know that, @ricardocvel could you try this way?

Coquinho:
[D12037](https://phab.enlightenment.org/D12037#inline-51953) address eina thread being treated as `uintptr` on native-windows too, so it probably resolves the problem with `ecore_threads`.

joaoantoniocardoso:
I am closing this in favor of #294

PR #126, opened 2020-06-08
Reverse eina_thread_name_set definition

Coquinho:
The current definition is causing an error at OSX build, as it doesn't have `pthread_set_name_np`.

### DIFF
--- a/src/lib/eina/eina_inline_thread_posix.x
+++ b/src/lib/eina/eina_inline_thread_posix.x
@@ -100,14 +100,27 @@ _eina_thread_join(Eina_Thread t)
 }
 
 static inline Eina_Bool
-_eina_thread_name_set(Eina_Thread thread, char *buf)
+_eina_thread_name_set(Eina_Thread thread, char *name)
 {
-#ifndef __linux__
-   pthread_set_name_np((pthread_t)t, buf);
+#ifdef EINA_HAVE_PTHREAD_SETNAME
+   char buf[16];
+   if (name)
+     {
+        strncpy(buf, name, 15);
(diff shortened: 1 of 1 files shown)


Coquinho:
This is a draft because I'm not confident with this approach, what do you guys think?
With this and other OSX PR, this [travis-ci](https://travis-ci.com/github/Coquinho/efl/jobs/346320328) builds. Every linux tests work as well.

src/lib/eina/eina_inline_thread_posix.x
```diff
@@ -100,14 +100,27 @@ _eina_thread_join(Eina_Thread t)
...
+#ifdef EINA_HAVE_PTHREAD_SETNAME
+   char buf[16];
+   if (name)
+     {
+        strncpy(buf, name, 15);
+        buf[15] = 0;
+     }
+   else buf[0] = 0;
+ #ifndef __linux__
+   pthread_set_name_np((pthread_t)thread, buf);
```
cauebs, line 114: Shouldn't it be like this?
```suggestion
   pthread_setname_np((pthread_t)thread, buf);
```

Coquinho: So this is o an `#ifndef __linux__` branch, I think it uses the [FreeBSD functions](https://www.freebsd.org/cgi/man.cgi?query=pthread_set_name_np&sektion=3&apropos=0&manpath=FreeBSD+7.1-RELEASE).

cauebs: Ok, got it.

PR #129, opened 2020-06-09
For now we should ignore libs in other builds too

Coquinho:
This PR deasble every lib that is not evil and eina for every plataform.


JPTIZ:
Could you explain the reason exactly?

Coquinho:
So we compile on linux

cauebs:
I think it makes sense. We can tackle problems on the linux build as we re-enable libs, one by one. CI will remind us.

Coquinho:
@walac and @JPTIZ convinced me that letting then disabled only at Windows builds would be better as we could be sure our changes didn't break any other thing (when Travis-CI is on).
So I'm closing this.

PR #131, opened 2020-06-09
windows: Remove Unposix completely

JPTIZ:
Removes unposix completely, thus ensuring that no unistd.h or sys/time.h is being included in windows.


src/lib/eina/eina_bezier.c
```diff
@@ -15,6 +15,8 @@
  * License along with this library;
  * if not, see <http://www.gnu.org/licenses/>.
  */
+#include <evil_private.h>
```
joaoantoniocardoso, line 18:
```suggestion
#ifdef _WIN32
# include <evil_private.h>
#endif
```

src/lib/eina/eina_pipe.h
```diff
@@ -1,17 +0,0 @@
-#ifndef EINA_PIPE_H
```
joaoantoniocardoso, line 1: Why removing this file?

src/tests/eina/eina_test_value.c
```diff
@@ -24,6 +24,14 @@
 #include <inttypes.h>
 #include <float.h>
 
+#ifdef _WIN32
+# ifndef WIN32_LEAN_AND_MEAN
+#  define WIN32_LEAN_AND_MEAN
+# endif
+# include <winsock2.h>
```
Coquinho, line 31: Wasn't this already passing? Why now we need `winsock2.h` here?

src/lib/eina/eina_main.c
```diff
@@ -27,7 +27,9 @@
 # ifdef HAVE_SYS_TYPES_H
 #  include <sys/types.h>
 # endif
-# include <unistd.h>
+# ifndef _WIN32
+#  include <unistd.h>
```
walac, line 31: Whouldn't this change hit cygwin builds?

JPTIZ: Indeed, didn't realize that. So I should change to `_MSC_VER`, right?

walac: Ideally, we should use a combination of `_WIN32` and `__CYGWIN__` macros, but I think `_MSC_VER` does the job.

src/lib/eina/eina_value.c
```diff
@@ -39,6 +48,8 @@
 #include "eina_safety_checks.h"
 #include "eina_value.h"
 
+#include <evil_private.h>
```
joaoantoniocardoso, line 51: shouldn't it go inside _WIN32 guards?

PR #135, opened 2020-06-09
Move EINA_THREAD_CLEANUP{_POP/_PUSH} to eina_thread_<arq>.h

Coquinho:
It seems that a `define` made at .x isn't available outside the file, which
caused some undefined reference at `ecore_thread.c`.
By moving it to `.h` it became visible.


Coquinho:
When approved I will squash and merge.

PR #136, opened 2020-06-10
Reenable some Linux functions

Coquinho:
Here I enable some Linux functions while maintaining them disabled at
`configure.bat` for windows.

### DIFF
--- a/src/generic/evas/meson.build
+++ b/src/generic/evas/meson.build
@@ -36,15 +36,11 @@ foreach loader : generic_loaders
   if get_option('evas-loaders-disabler').contains(loader) == false
     subdir(loader)
 
-    evas_dep = [eina, generic_deps]
-    if sys_windows
-      evas_dep += [evil_unposix]
-    endif
     tmp = executable('evas_image_loader.'+loader,
         generic_src,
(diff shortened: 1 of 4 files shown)


src/generic/evas/meson.build
```diff
@@ -36,11 +36,15 @@ foreach loader : generic_loaders
   if get_option('evas-loaders-disabler').contains(loader) == false
     subdir(loader)
 
+    evas_dep = [eina, generic_deps]
+    if sys_windows
+      evas_dep += [evil_unposix]
```
walac, line 41: Is this still relevant after the unposix patch?

Coquinho: After the unposix patch, it will not be relevant, but as it isn't at `native-windows` yet, we need it.
If you guys prefer, I could make the patch depending on #131, do you think it would be better @walac, @joaoantoniocardoso?

walac: Either that or you could create an issue to remove this piece of code once #131 lands. Both approaches are fine by me.

Coquinho:
This has been rebased, please, have a look again

PR #139, opened 2020-06-10
[Deps] Ector

joaoantoniocardoso:
After eet (#138), ector is building with no effort.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -515,7 +515,6 @@ if sys_windows
     'ecore_win32',
     'ecore_wl2',
     'ecore_x',
-    'ector',
     'edje',
     'eeze',
     'efl_canvas_wl',
--- a/meson.build
+++ b/meson.build
@@ -517,7 +517,6 @@ if sys_windows
     'ecore_x',
     'ector',
     'edje',
-    'eet',
     'eeze',
     'efl_canvas_wl',
     'efreet',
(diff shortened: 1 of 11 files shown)


win32-deps/meson.build
```diff
@@ -1,17 +1,5 @@
...
-  freetype_lib_dir = 'freetype.2.8.0.1/build/native/lib/x64/v141/static/Release'
-  freetype_lib = cc.find_library('freetype28',
-    dirs: meson.build_root() + '/../win32-deps/' + freetype_lib_dir,
-    required: true,
-  )
-  freetype = declare_dependency(
-    dependencies: freetype_lib,
-    include_directories: freetype_inc_dir,
-  )
-
```
Coquinho, line 14: Why removing this?

joaoantoniocardoso: It was wrongly calling `freetype2` as `freetype`, which for efl is built from `static_libs/freetype`. At this time I didn't realize that and just removed. Later on `Evas` I revert this one and did a proper fix (23f5e21 and f49f6c3).

Since Evas PR will fix this I think it is unnecessary to apply those two commits here too.

Coquinho: Things are in 2 commits here, why not removing `eb6a1147605e464c61fed33b9d3abcbc726d50c6` that does this?

PR #144, opened 2020-06-15
Linux build: Explicitly add ecore_x dependency to elementary examples

Coquinho:
For some reason, `ecore_x` wasn't being set at `elementary` or at least while compiling `win_example.c`, so this PR explicitly adds `ecore_x`.

### DIFF
--- a/src/examples/elementary/meson.build
+++ b/src/examples/elementary/meson.build
@@ -123,8 +123,15 @@ examples = [
   'efl_canvas_textblock_obstacles_example'
 ]
 
+examples_deps = [elementary, ecore, eio, m]
+
+if get_option('x11')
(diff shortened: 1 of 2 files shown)


Coquinho:
So I made the PR, but no 100% sure about this solution, as this is not how `master` handles it and by some debugging is easy to see that at `src/lib/elementary/meson.build` `ecore_x` is added to `elementary` variable, even though it doesn't show at [compilation](https://travis-ci.com/github/Coquinho/efl/jobs/348365330#L5050).

Coquinho:
Moving it to draft as I'm not sure about it.

Coquinho:
As discussed in #260 I'm making this ready again.

PR #146, opened 2020-06-16
Add files to eina public_headers

Coquinho:
`eina_inline_lock_posix.x`, `eina_inline_lock_win32.x` and `eina_pipe.h`
weren't available as `eina` `public_sub_headers` at `meson.build` making
then unavailable when instaling our build.

### DIFF
--- a/src/lib/eina/meson.build
+++ b/src/lib/eina/meson.build
@@ -114,6 +114,7 @@ public_sub_headers = [
 'eina_vpath.h',
 'eina_abstract_content.h',
 'eina_fnmatch.h',
+'eina_pipe.h',
 ]
 
 public_headers = [
@@ -207,6 +208,7 @@ if sys_windows
   )
   public_sub_headers += files(
     'eina_thread_win32.h',
+    'eina_inline_lock_win32.x',
   )
 else
   eina_src += files(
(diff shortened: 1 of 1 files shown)


walac:
I believe those files shouldn't be in public headers. Their include directives should be in the eina implementation files.

Coquinho:
> I believe those files shouldn't be in public headers. Their include directives should be in the eina implementation files.

Didn't get it, @walac. What do you mean by implementation files? The `eina_src`?

walac:
> Didn't get it, @walac. What do you mean by implementation files? The eina_src?

.c files

Coquinho:
But they are not .c files, that's why they went to `public_headers`, like every other inline file on `eina`

JPTIZ:
> I believe those files shouldn't be in public headers. Their include directives should be in the eina implementation files.

The problem is that some definitions from public headers (e.g. `Eina_Thread` in `eina_thread.h`) are platform dependant, thus requiring that the definitions on these `.x` files be included in these headers. What do you think that could be a good workaround?

walac:
> The problem is that some definitions from public headers (e.g. Eina_Thread in eina_thread.h) are platform dependant, thus requiring that the definitions on these .x files be included in these headers. What do you think that could be a good workaround?

Ah, I see. The the `.x` files are part of public headers in upstream, then I see no problem including them here too.

PR #147, opened 2020-06-16
Getting freetype2 from meson wrap

joaoantoniocardoso:
Currently freetype2 is get using Nuget, but as freetype2 do exists as a meson wrap, we could be using it to built from sources instead of getting binaries with Nuget.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,8 +100,5 @@ efl_libs.csv
 env.bat
 
 # Subprojects
-subprojects/pcre-*
-subprojects/zlib-*
-subprojects/getopt
-subprojects/packagecache
-subprojects/check*
+subprojects/*
+!subprojects/*.wrap
--- a/header_checks/meson.build
+++ b/header_checks/meson.build
@@ -234,6 +234,11 @@ if sys_windows
 
   # getopt
   getopt = subproject('getopt').get_variable('getopt_dep')
+
+  # freetype2
+  freetype2_project = subproject('freetype2')
+  freetype2 = freetype2_project.get_variable('libfreetype')
+  freetype2_dep = freetype2_project.get_variable('freetype_dep')
 else
   zlib = dependency('zlib')
   if not cc.has_header_symbol('regex.h', 'regcomp')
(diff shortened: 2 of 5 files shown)


JPTIZ:
Why did we stop this exactly?

joaoantoniocardoso:
> Why did we stop this exactly?

Actually I don't remember very well, I think at the time it was only building in my machine for unknown reasons... Maybe something related with the recent patches about pkgconfig?

.gitignore
```diff
@@ -100,8 +100,5 @@ efl_libs.csv
 env.bat
 
 # Subprojects
-subprojects/pcre-*
-subprojects/zlib-*
-subprojects/getopt
-subprojects/packagecache
-subprojects/check*
+subprojects/*
+!subprojects/*.wrap
```
JPTIZ, line 104: You're a genious!

joaoantoniocardoso:
@JPTIZ Note that the dependency called `freetype` is not equivalent to `freetype2`. The dependency `freetype` is located at `src/static_libs/freetype` and those who wants `freetype` is requesting headers that doesn't exist in `freetype2`.

JPTIZ:
> @JPTIZ Note that the dependency called `freetype` is not equivalent to `freetype2`. The dependency `freetype` is located at `src/static_libs/freetype` and those who wants `freetype` is requesting headers that doesn't exist in `freetype2`.

Now things do make sense, thanks haha

header_checks/meson.build
```diff
@@ -219,6 +219,9 @@ endif
 # libcheck
 check = dependency('check', fallback : ['check', 'check_dep'])
 
+# freetype2
+freetype2 = dependency('freetype2', fallback : ['freetype2', 'freetype_dep'])
```
joaoantoniocardoso, line 223: shouldn't it be `freetype2_dep`?

JPTIZ: If you're saying it should be `freetype2_dep = [...]`: nope, `freetype2` is the variable used around EFL (while `freetype2_dep` has no results). 

If you're saying it should be `['freetype2', 'freetype2_dep']`: also nope, [see the wrap's meson variable](https://github.com/mesonbuild/freetype2/blob/2.9.1/meson.build#L227).

joaoantoniocardoso: you're right  =)

PR #151, opened 2020-06-16
windows: Implement ftruncate

cauebs:
Also reenables evil tests and implements a test case for `ftruncate`.
I noticed that `evil_test_dlfcn` is failing and I'll try to fix it in a subsequent PR.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -437,7 +437,7 @@ endif
 
 subprojects = [
 # name              |   option              | mod  | lib  | bin  | bench | tests | examples | true if build in efl-one | pkg-config options | name of static libs
-['evil'             ,[]                    , false,  true, false, false, false, false,  true, [], []],
+['evil'             ,[]                    , false,  true, false, false,  true, false,  true, [], []],
 ['eina'             ,[]                    , false,  true,  true,  true,  true,  true,  true, [], []],
 ['eolian'           ,[]                    , false,  true,  true, false,  true, false, false, ['eina'], []],
 ['eo'               ,[]                    , false,  true, false,  true,  true, false,  true, ['eina'], []],
--- a/src/lib/evil/evil_unistd.c
+++ b/src/lib/evil/evil_unistd.c
@@ -23,7 +23,7 @@ LONGLONG _evil_time_freq;
 LONGLONG _evil_time_count;
 long     _evil_time_second;
 
-long 
+long
 _evil_systemtime_to_time(SYSTEMTIME st);
 
 long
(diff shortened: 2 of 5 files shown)


Coquinho:
This seems good to send directly to phab, @cauebs .

Coquinho:
How can I test this PR?

cauebs:
> How can I test this PR?

Unsure. It's not used all that much. We could add a test case to `evil_test_unistd.c`.

But now that I'm looking at the call sites, it seems EFL is using it under the incorrect assumption that only negative return values signal an error. Both error codes I'm using in the current implementation are positive, loosely mirroring what I found in the Linux manpage for this function. Now, do I fix the usage or "fix" the implementation?

Coquinho:
> > How can I test this PR?
> 
> Unsure. It's not used all that much. We could add a test case to `evil_test_unistd.c`.

If you know how to test it, please, do a test case.

> But now that I'm looking at the call sites, it seems EFL is using it under the incorrect assumption that only negative return values signal an error. Both error codes I'm using in the current implementation are positive, loosely mirroring what I found in the Linux manpage for this function. Now, do I fix the usage or "fix" the implementation?

@cauebs, what do you think would be faster for us? I think that changing how efl uses it would need too much effort, we would probably need a phab diff to first change how they use it and another one to add the correct implementation + tests.

...But having a longer thought, there should be no problem on opening the diff on phab, but already using it in `native-windows`, what you guys think? I wouldn't mind the opening of those too diffs while already using it here.

cauebs:
Oh, no, I was mistaken. These error numbers are set to errno. The return value is indeed 0 if successful, otherwise -1. Is there a way to reproduce this in EFL? Should I use SetLastError on Windows? But then, with which values, POSIX or some Windows close equivalents? Or should I ignore these and just return -1?

Coquinho:
> Oh, no, I was mistaken. These error numbers are set to errno. The return value is indeed 0 if successful, otherwise -1. Is there a way to reproduce this in EFL? Should I use SetLastError on Windows? But then, with which values, POSIX or some Windows close equivalents? Or should I ignore these and just return -1?

@cauebs which error values we need? Maybe we can use [ucrt errno](https://docs.microsoft.com/en-us/cpp/c-runtime-library/errno-constants?view=vs-2019) if all needed error codes are available?

cauebs:
I just noticed my editor deleted some trailing whitespace. Is that a problem?

JPTIZ:
> I just noticed my editor deleted some trailing whitespace. Is that a problem?

Tbh not really, just people don't like messing whitespace changes with ""real code"" changes. Specially if it leads to minor unnecessary conflicts.

cauebs:
If you guys prefer it, I could try and fix the failing test (`evil_test_dlfcn`) in this PR, but it seems to me that it's a bit out of scope. There are already other failing tests enabled (eo and eolian), so I don't consider it a significant regression.

JPTIZ:
> If you guys prefer it, I could try and fix the failing test (`evil_test_dlfcn`) in this PR, but it seems to me that it's a bit out of scope. There are already other failing tests enabled (eo and eolian), so I don't consider it a significant regression.

Another PR seems to make more sense for me. Reenabling tests here make sense - yeah. But tackling another test whose errors you uncovered by reenabling tests? That's two different issues.

joaoantoniocardoso:
@cauebs could you recreate this PR but for native-windows?

PR #152, opened 2020-06-16
Reenable eina file tests

walac:


src/tests/eina/eina_test_file.c
```diff
@@ -491,7 +491,7 @@ static void *
 _eina_test_file_thread(void *data EINA_UNUSED, Eina_Thread t EINA_UNUSED)
 {
 #ifdef _WIN32
-   const char *filename = "cmd.exe";
+   const char *filename = "C:/Windows/system.ini";
```
Coquinho, line 494: What is this file?

walac: Just a test file with an absolute path for test

PR #156, opened 2020-06-18
Link ecore_x to elementary examples

walac:
ecore_x was missing in linking dependencies.


walac:
Dup of #144. Closing.

PR #161, opened 2020-06-19
Devs/felipealmeida/native windows/ecore api

felipealmeida:


joaoantoniocardoso:
I have some doubts here:
- Why the symbols `EWAPI` and `EOAPI` from `src/tests/ecore` was changed to `EAPI EAPI_WEAK` and these same symbols from `src/libs/ecore` was changed to `ECORE_API ECORE_API_WEAK`?
- The symbol `ECORE_STATIC` shouldn't be defined in meson.build, like `ECORE_BUILD`?

JPTIZ:
> I have some doubts here:
> 
>     * Why the symbols `EWAPI` and `EOAPI` from `src/tests/ecore` was changed to `EAPI EAPI_WEAK` and these same symbols from `src/libs/ecore` was changed to `ECORE_API ECORE_API_WEAK`?

Good point. Probably both should be `ECORE_API ECORE_API_WEAK`.

>     * The symbol `ECORE_STATIC` shouldn't be defined in meson.build, like `ECORE_BUILD`?

`ECORE_STATIC` is probably going to be dropped actually, since static linking for EFL will be dropped.

Coquinho:
May we close this? #175 + #171 seems to do this.

walac:
> May we close this? #175 + #171 seems to do this.

 Closing

PR #162, opened 2020-06-19
Declare ignored subprojects as empty dependencies

Coquinho:
Some libs may try to use libs that are being ignored those causing
an undefined symbol at `configure.bat`, by declaring them as empty
dependencies this isn't a problem.


joaoantoniocardoso:
I don't know if I agree with this one. I think it will make harder to find what depends on what on our trail to enable one lib at each step.

Coquinho:
@joaoantoniocardoso ok, it makes sense, closing this.

PR #164, opened 2020-06-22
Add EINA_API to fnmatch

Coquinho:
`EINA_API` macro wasn't being used at eina_fnmatch


Coquinho:
@joaoantoniocardoso Oh, thanks, I was trying to use with `eina_api` instead for some reason...

PR #165, opened 2020-06-22
eina_api.h is no longer necessary

joaoantoniocardoso:
EINA_API is now defined in `src/lib/eina/eina_types.h` and `src/lib/eina/eina_api.h` is no longer necessary.


Coquinho:
This may need a rebase now.

Coquinho:
@felipealmeida maybe it would be a good idea to add this change to [D11993](https://phab.enlightenment.org/D11993)?

joaoantoniocardoso:
> @felipealmeida maybe it would be a good idea to add this change to [D11993](https://phab.enlightenment.org/D11993)?

eina_api.h doesn't exist in the upstream: [check here](https://github.com/Enlightenment/efl/blob/master/src/lib/eina/eina_api.h)

felipealmeida:
> 
> 
> @felipealmeida maybe it would be a good idea to add this change to [D11993](https://phab.enlightenment.org/D11993)?

Com certeza. Eu percebi que esse arquivo ainda existia ontem

JPTIZ:
> @felipealmeida maybe it would be a good idea to add this change to [D11993](https://phab.enlightenment.org/D11993)?

But...the file doesn't even exists on that diff.

PR #166, opened 2020-06-23
Changing check to check_dep

joaoantoniocardoso:
This patch should fix all the errors `ERROR: Unknown variable "check"`.

### DIFF
--- a/src/bindings/mono/meson.build
+++ b/src/bindings/mono/meson.build
@@ -205,7 +205,7 @@ if (get_option('dotnet'))
 
   lib_csproj_conf_data = configuration_data()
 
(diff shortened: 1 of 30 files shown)


JPTIZ:
I recommend waiting until I submit the subprojects diff to phab, so people can beat me for adding to every external dep the `_dep` suffix.

JPTIZ:
Tbh, we should do the opposite: declare `check` as `check_dep`.

I'll submit a new PR fixing this.

joaoantoniocardoso:
@JPTIZ [check this commit.](https://github.com/expertisesolutions/efl/pull/109/commits/a3c7f23a736ad5adb9975377aa36ba652ad8b2dc)

PR #168, opened 2020-06-23
unistd imports fix

joaoantoniocardoso:
There are several unistd imports left in the sources that needed to be fixed in order to be built on windows.

### DIFF
--- a/src/lib/ecore/ecore_anim.c
+++ b/src/lib/ecore/ecore_anim.c
@@ -4,20 +4,21 @@
 
 #include <stdlib.h>
 #include <math.h>
-#include <unistd.h>
 #include <string.h>
-#include <sys/types.h>
-#include <sys/stat.h>
+
+#ifndef _WIN32
+# include <unistd.h>
+# include <sys/types.h>
+# include <sys/stat.h>
+#else
+# include <evil_private.h>
+#endif
 
 #if defined(HAVE_SYS_EPOLL_H) && defined(HAVE_SYS_TIMERFD_H)
 # define HAVE_EPOLL   1
 # include <sys/epoll.h>
 # include <sys/timerfd.h>
 #endif
-
-#ifdef _WIN32
-#include <evil_private.h>
-#endif /* _WIN32 */
 
 #ifdef HAVE_PRCTL
 # include <sys/prctl.h>
--- a/src/lib/ecore/ecore_anim.c
+++ b/src/lib/ecore/ecore_anim.c
@@ -6,10 +6,11 @@
 #include <math.h>
 #include <string.h>
 
+#include <sys/types.h>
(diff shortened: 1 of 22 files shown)


JPTIZ:
I thought I handled them in #149.

joaoantoniocardoso:
> I thought I handled them in #149.

Possibly yes, do you have a commit or PR that isolates your unistd fix so I can use in other branches instead of using this one?

Coquinho:
@JPTIZ partially, not all changes done here, happen at #149, `eet` for example.

src/lib/ecore/ecore_anim.c
```diff
@@ -4,21 +4,22 @@
...
 #include <math.h>
-#include <unistd.h>
 #include <string.h>
-#include <sys/types.h>
-#include <sys/stat.h>
+
+#ifndef _WIN32
+# include <unistd.h>
+# include <sys/types.h>
+# include <sys/stat.h>
```
Coquinho, line 12: `ucrt` has `sys/types.h` and `sys/stat.h`

src/lib/ecore/ecore_time.c
```diff
@@ -5,7 +5,11 @@
 #define EFL_LOOP_PROTECTED
 
 #include <stdlib.h>
-#include <sys/time.h>
+#ifndef _WIN32
+# include <sys/time.h>
+#else
+# include <evil_private.h>
```
Coquinho, line 11: why evil is needed here?

src/lib/ecore/efl_thread.c
```diff
@@ -7,17 +7,22 @@
 #define EFL_IO_CLOSER_PROTECTED 1
 
 #ifdef _WIN32
-# include <evil_private.h> /* pipe fcntl */
```
Coquinho, line 10: those guards are still here.

src/lib/ecore/ecore_main.c
```diff
@@ -13,12 +13,17 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
-#include <unistd.h>
 #include <math.h>
-#include <sys/types.h>
 #include <errno.h>
+#include <sys/types.h>
```
Coquinho, line 18: I think we should let `sys/types` where it was, I don't know if phab will like this latter.

src/lib/ecore/ecore_main_timechanges.c
```diff
@@ -5,12 +5,17 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
-#include <unistd.h>
 #include <math.h>
-#include <sys/types.h>
```
Coquinho, line 10: Here `sys/types` is still removed

PR #159, opened 2020-06-18
 implementation of gettimeofday for windows with GetSystemTime

ricardocvel:
Define `evil_gettimeofday`, if `HAVE_GETTIMEOFDAY` is not defined, do it
and define `gettimeofday` as `evil_gettimeofday`.

PR #169, opened 2020-06-23
[Evil] EAPI -> EVIL_API

joaoantoniocardoso:


felipealmeida:
Tem como fazer um squash antes de mergear/rebase/cherry-pick ? Não faz sentido ficar com 2 commits

joaoantoniocardoso:
> Tem como fazer um squash antes de mergear/rebase/cherry-pick ? Não faz sentido ficar com 2 commits

Sim, a gente tem usado **_squash and merge_** .

src/lib/evil/evil_api.h
```diff
@@ -1,25 +1,25 @@
 #ifndef EVIL_API_H
 #define EVIL_API_H
 
-#ifdef EAPI
-#undef EAPI
+#ifdef EVIL_API
+#undef EVIL_API
```
Coquinho, line 5: If `EVIL_API` is already defined shouldn't it yield an error? As it shouldn't be already defined?

joaoantoniocardoso:
> Shouldn't this `evil_api.h` use `EVIL_STATIC` and `EVIL_BUILD`?

Yeah, you're right, I'll change it to work in the same way of the other APIs.

joaoantoniocardoso:
Writing just to remember that #59 and #159 is adding new functions that this patch is not seeing yet.

src/lib/evil/evil_api.h
```diff
@@ -1,26 +1,44 @@
 #ifndef EVIL_API_H
 #define EVIL_API_H
 
-#ifdef EAPI
-#undef EAPI
+#ifdef EVIL_API
+#error EVIL_API should not be already defined
 #endif
```
JPTIZ, line 6: If we're using `evil_api.h`, there's no reason to check this at all actually, since the include-guards will handle it.

PR #170, opened 2020-06-23
ecore_x: Change EAPI symbols for ECORE_X_API

Coquinho:


Coquinho:
@joaoantoniocardoso it should be like this?

src/lib/ecore_x/ecore_x_api.h
```diff
@@ -0,0 +1,34 @@
...
+# ifndef ECORE_X_STATIC
+#  ifdef ECORE_X_BUILD
+#   define ECORE_X_API __declspec(dllexport)
+#  else
+#   define ECORE_X_API __declspec(dllimport)
+#  endif
+# else
+#  define ECORE_X_API
+# endif
+# define ECORE_X_API_WEAK
```
Coquinho, line 18: @felipealmeida do we need a `weak` at `ecore_x`? This seems to be only used here at this file.

JPTIZ: And there's no `WAPI` on the original (master's) code.

Coquinho:
@JPTIZ this may be changed later

PR #171, opened 2020-06-23
[Eo + Eolian] Change EAPI symbols for EO_API and EOLIAN_API

Coquinho:


joaoantoniocardoso:
Actually this one could benefit from [847eb43](https://github.com/expertisesolutions/efl/pull/109/commits/847eb43f3da744ff349ce5d1d96acd7559539f86).

src/tests/eolian/data/class_simple_ref.c
```diff
@@ -29,7 +29,7 @@ __eolian_class_simple_a_get_reflect(const Eo *obj)
    return eina_value_int_init(val);
 }
 
-EOAPI EFL_FUNC_BODY_CONST(efl_canvas_object_simple_a_get, int, 100);
+EAPI EAPI_WEAK EFL_FUNC_BODY_CONST(efl_canvas_object_simple_a_get, int, 100);
```
Coquinho, line 32: @felipealmeida why using `EAPI EAPI_WEAK` here? Shouldn't it be an `EO_API` or something like that?

src/lib/eo/eo_api.h
```diff
@@ -0,0 +1,36 @@
...
+#define _EFL_EO_API_H
+
+#ifdef EO_API
+#error EO_API should not be already defined
+#endif
+
+#define EOLIAN
+
+#ifdef _WIN32
+# ifndef EO_STATIC
```
Coquinho, line 11: @felipealmeida, It seams that `EO_STATIC` is never defined, where and how it should be settled to build Eo statically?
And another thing isn't this letting `EFL one` out?

src/tests/eolian/data/class_simple_ref.c
joaoantoniocardoso: Yeah I think we should use `EO_API EO_API_WEAK` here.

Coquinho: Actually, as this is a test, we don't need any API setting, as it will only be used inside itself, it will not be exported/imported.

Coquinho: Done

src/tests/eolian/data/class_simple_ref.c
```diff
@@ -17,7 +17,7 @@ __eolian_class_simple_a_set_reflect(Eo *obj, Eina_Value val)
    return r;
 }
 
-EAPI EAPI_WEAK EFL_FUNC_BODYV(efl_canvas_object_simple_a_set, Eina_Bool, EINA_TRUE /* true */, EFL_FUNC_CALL(value), int value);
+ EFL_FUNC_BODYV(efl_canvas_object_simple_a_set, Eina_Bool, EINA_TRUE /* true */, EFL_FUNC_CALL(value), int value);
```
joaoantoniocardoso, line 20: There is a blank space left in every `EAPI EAPI_WEAK` replacement in the last commit.

Coquinho: oh boy....

Coquinho: done, thanks

Coquinho:
Hey @walac, there were changes after your approval, so please look again.

PR #172, opened 2020-06-23
[Eina c++] tests: Change EAPI symbols for eina C++ tests

Coquinho:


src/tests/eina_cxx/eina_cxx_suite.h
```diff
@@ -19,4 +19,7 @@ void eina_test_optional(TCase* tc);
 void eina_test_value(TCase* tc);
 void eina_test_log(TCase* tc);
 
+#define EINACXXTEST_API
```
Coquinho, line 22: @felipe this seems to never be used, it should be like this?

Coquinho: oh, thanks @JPTIZ

PR #173, opened 2020-06-23
[efl] Change EAPI symbols for EFL_API

Coquinho:


Coquinho:
Hey @walac, there were changes after your approval, so please look again.

PR #175, opened 2020-06-23
[ecore] Rename EAPI symbol in ecore to ECORE_API

Coquinho:


src/lib/ecore/ecore_private.h
```diff
@@ -247,6 +222,12 @@ EAPI void ecore_evas_object_animator_init(Ecore_Evas_Object_Animator_Interface *
 typedef unsigned int Ecore_Magic;
 #define ECORE_MAGIC               Ecore_Magic __magic
 
+ECORE_API void
+_ecore_magic_fail(const void *d,
+                  Ecore_Magic m,
+                  Ecore_Magic req_m,
+                  const char *fname EINA_UNUSED);
+
```
Coquinho, line 230: #L267 has a definition of `_ecore_magic_fail` too.

joaoantoniocardoso: It seems a mistake, maybe add the `EINA_UNUSED` to the other and remove this one?

Coquinho: I have no idea because the one ate line 267 is from `master`, and  desires that I don't know what we would need to set `fname` as `EINA_UNUSED`  for windows build

Coquinho:
@walac and @joaoantoniocardoso there were changes here too, please take a look.

Do you guys have any idea about [this](https://github.com/expertisesolutions/efl/pull/175#discussion_r445093316)?

PR #176, opened 2020-06-23
[eldbus] Change EAPI symbols for ELDBUS_API

Coquinho:


src/lib/eldbus/eldbus_api.h
```diff
@@ -0,0 +1,34 @@
...
+#  endif
+# else
+#  define ELDBUS_API
+# endif
+# define ELDBUS_API_WEAK
+#else
+# ifdef __GNUC__
+#  if __GNUC__ >= 4
+#   define ELDBUS_API __attribute__ ((visibility("default")))
+#   define ELDBUS_API_WEAK __attribute__ ((weak))
```
Coquinho, line 23: weak don't seems to be used.

Coquinho:
> I think we need to add `-DELDBUS_BUILD` to `src/lib/eldbus/meson.build:102`.

@joaoantoniocardoso Done.

src/lib/eldbus/eldbus_api.h
joaoantoniocardoso: It seems that we are defining _weak_ for all libs and not necessarily using them

Coquinho: yeah, I don't know if people at phab will like that.
Should we just do it like this for now, and later update it with phabs?

joaoantoniocardoso: Yeah I think. It will not be hard to remove definitions from all APIs if necessary

Coquinho:
Hey @joaoantoniocardoso, @walac, there were changes after your approval, so please look again.

PR #177, opened 2020-06-23
[eio] Change EAPI symbols for EIO_API

Coquinho:


Coquinho:
> I think we need to add `-DEIO_BUILD` to `src/lib/eio/meson.build:76`.

Done, @joaoantoniocardoso

PR #180, opened 2020-06-23
[modules] Rename EAPI symbol to EMODAPI symbol

Coquinho:
Differentiate building a module and building a EFL library

### DIFF
--- a/src/modules/ecore_evas/engines/drm/ecore_evas_drm.c
+++ b/src/modules/ecore_evas/engines/drm/ecore_evas_drm.c
@@ -30,14 +30,14 @@
 
 #ifdef _WIN32
 # ifdef DLL_EXPORT
-#  define EAPI __declspec(dllexport)
+#  define EMODAPI __declspec(dllexport)
 # else
 #  define EAPI
 # endif /* ! DLL_EXPORT */
 #else
 # ifdef __GNUC__
 #  if __GNUC__ >= 4
-#   define EAPI __attribute__ ((visibility("default")))
+#   define EMODAPI __attribute__ ((visibility("default")))
 #  else
 #   define EAPI
 #  endif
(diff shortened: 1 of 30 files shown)


src/modules/ecore_evas/engines/drm/ecore_evas_drm.c
```diff
@@ -30,14 +30,14 @@
 
 #ifdef _WIN32
 # ifdef DLL_EXPORT
-#  define EAPI __declspec(dllexport)
+#  define EMODAPI __declspec(dllexport)
```
Coquinho, line 33: I don't know how modules work, is this correct?

src/modules/evas/engines/gl_common/evas_gl_core.h
```diff
@@ -8,14 +8,10 @@
...
-#  ifdef DLL_EXPORT
-#   define EMODAPI __declspec(dllexport)
-#  else
-#   define EMODAPI
-#  endif
+# ifndef EFL_MODULE_STATIC
+#  define EMODAPI __declspec(dllexport)
 # else
-#  define EMODAPI __declspec(dllimport)
+#  define EMODAPI
```
JPTIZ: Wait: we're removing the `dllimport` possibility. Is this right (I didn't work too much with modules, so I sincerely don't know)?

Coquinho: @JPTIZ this was discussed in [slack](https://expertisesolutions.slack.com/archives/C7NGCQY8M/p1593524654020200?thread_ts=1593524654.020200).

JPTIZ: Thanks. Modules are never imported since they're loaded manually in runtime, so it makes sense.

PR #183, opened 2020-06-23
[elementary] Change EAPI symbol for elementary and its tests

Coquinho:


src/lib/elementary/Elementary.h
```diff
@@ -89,18 +91,18 @@
 #ifdef _WIN32
 # ifdef EFL_BUILD
```
joaoantoniocardoso, line 92: Should it use `EFL_BUILD`? I think this whole EAPI thing here is already defined at the above included `elementary_api.h`.

Coquinho: This probably should include `elementary_api.h` instead of redefining it.

src/lib/elementary/elc_scrolled_entry.c
```diff
@@ -4,194 +4,194 @@
 
 #include <Elementary.h>
 #include "elm_priv.h"
-EINA_DEPRECATED EAPI Evas_Object *
+EINA_DEPRECATED ELM_API Evas_Object *
```
Coquinho, line 7: This should follow https://docs.microsoft.com/pt-br/cpp/build/exporting-from-a-dll-using-declspec-dllexport?view=vs-2017?

Coquinho: I mean, this part specifically:
> To export functions, the __declspec(dllexport) keyword must appear to the left of the calling-convention keyword, if a keyword is specified.

JPTIZ: In this case, this is not necessary: `EINA_DEPRECATED` is also a `__declspec`.

Coquinho:
Hey @joaoantoniocardoso, @walac, @JPTIZ, there were changes, so please look again.

PR #184, opened 2020-06-23
[emotion] Change EAPI symbols for sublib EMOTION_API

Coquinho:


src/lib/emotion/emotion_modules.h
```diff
@@ -10,17 +10,17 @@
 #ifdef _WIN32
 # ifdef EFL_BUILD
 #  ifdef DLL_EXPORT
-#   define EAPI __declspec(dllexport)
+#   define EMOTION_API __declspec(dllexport)
```
Coquinho, line 13: maybe instead of redefining `EMOTION_API` this should include `emotion_api.h`?

Coquinho:
There were changes, so please look again.

PR #185, opened 2020-06-23
[evas] Change EAPI symbols for EVAS_API

Coquinho:


src/lib/evas/include/evas_common_private.h
```diff
@@ -1151,7 +1151,7 @@ EVAS_API void generic_cache_data_drop(Generic_Cache *cache, void *key);
 }
 #endif
 
-#undef EAPI
-#define EAPI
+#undef EVAS_API
+#define EVAS_API
```
joaoantoniocardoso, line 1155: Is it really necessary?

src/lib/evas/include/evas_private.h
```diff
@@ -1577,8 +1552,8 @@ _gfx_to_evas_render_op(Efl_Gfx_Render_Op rop)
 }
 #endif
 
-#undef EAPI
-#define EAPI
+#undef EVAS_API
```
joaoantoniocardoso, line 1555: I think this _undef define_ should be removed to the ensure the same definition for `EVAS_API`  to be used in the files that is using this one as headers.

src/lib/evas/include/evas_common_private.h
Coquinho: O.o

Coquinho: working on it

Coquinho:
There were changes, so please look again.

PR #188, opened 2020-06-25
Reenable evil suite

### LINKED ISSUE #160: evil_test_dlfcn is failing
In #151 I reenabled tests for evil and noticed that this test case was already failing.
I'm just filing this so I don't forget to fix that.

cauebs:
Closes  #160.

### DIFF
--- a/src/tests/evil/meson.build
+++ b/src/tests/evil/meson.build
@@ -1,0 +1,24 @@
+evil_test_src = files(
+'evil_suite.c',
+'evil_suite.h',
+'evil_test_dlfcn.c',
+'evil_test_main.c',
+'evil_test_stdio.c',
+'evil_test_stdlib.c',
+'evil_test_unistd.c',
+)
+
+evil_test_exe = executable('evil_suite',
+  include_directories : config_dir,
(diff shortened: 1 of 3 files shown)


cauebs:
Well, yeah. As I mentioned on our slack server, the evil test suite is pretty slow, and at first glance it seems that it's really the evil functions that are slow. I think I'll open an issue so we take a closer look some day.

src/tests/evil/meson.build
```diff
@@ -0,0 +1,24 @@
...
+'evil_test_main.c',
+'evil_test_stdio.c',
+'evil_test_stdlib.c',
+'evil_test_unistd.c',
+)
+
+evil_test_exe = executable('evil_suite',
+  include_directories : config_dir,
+  sources : evil_test_src,
+  dependencies: [m, check_dep, eina, evil],
```
Coquinho, line 14: this tests depend on eina?

cauebs: Yes, it does. `evil_suite.h` includes `efl_check.h`, which in turn includes `Eina.h`. But great catch nonetheless, it does seem strange.

PR #189, opened 2020-06-25
build: Use PKG_CONFIG_LIBDIR to ignore pkg-config from strawberry-perl

JPTIZ:
This will probably make some new breaking changes, e.g. `libjpeg` not being found anymore :) But it is for good.

### DIFF
--- a/configure.bat
+++ b/configure.bat
@@ -36,6 +36,8 @@ exit /B %errorlevel%
         exit /B 1
     )
     set all_set=
+
+    set pkg_config_libdir=thistotallydoesnotexist
 exit /B 0
 
 


joaoantoniocardoso:
Why to not use strawberry's shipped libs for our dependencies in ci ?

Coquinho:
@JPTIZ Shouldn't this PR remove the `if sys_windows` like [here](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/meson.build#L11) and any other guarding `pkgconfig`?

JPTIZ:
> @JPTIZ Shouldn't this PR remove the `if sys_windows` like [here](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/meson.build#L11) and any other guarding `pkgconfig`?

This is a thing for another PR :D For now we're just ignoring **where pkg-config searches for dependencies**, not anything related to pkg-config file generation (which is why that `pkg_config` variable exists).

JPTIZ:
> Why to not use strawberry's shipped libs for our dependencies in ci ?
> 
> Also, probably this change will not be a problem as `libjpeg` seems to be working fine with Nuget enabled.

Because some of them can be a "lie" :D Like libjpeg itself, that here it is not installed and it is not part of our build-system neither from default Windows installation - instead, it comes from strawberry-pie. You can think of your question as analogue to "why not using msys?", to which the answer is "because that's not what we're aiming here".

joaoantoniocardoso:
> > Why to not use strawberry's shipped libs for our dependencies in ci ?
> > Also, probably this change will not be a problem as `libjpeg` seems to be working fine with Nuget enabled.
> 
> Because some of them can be a "lie" :D Like libjpeg itself, that here it is not installed and it is not part of our build-system neither from default Windows installation - instead, it comes from strawberry-perl. You can think of your question as analogue to "why not using msys?", to which the answer is "because that's not what we're aiming here".

In my toughs it would make sense to automatically get dependencies from whatever corner from the system if the lib met some requirements that should be well expressed in meson.build (checking version, headers, symbols).
Are we aiming to manage the distribution of every dependency for native windows?

JPTIZ:
> > > Why to not use strawberry's shipped libs for our dependencies in ci ?
> > > Also, probably this change will not be a problem as `libjpeg` seems to be working fine with Nuget enabled.
> > 
> > 
> > Because some of them can be a "lie" :D Like libjpeg itself, that here it is not installed and it is not part of our build-system neither from default Windows installation - instead, it comes from strawberry-perl. You can think of your question as analogue to "why not using msys?", to which the answer is "because that's not what we're aiming here".
> 
> In my toughs it would make sense to automatically get dependencies from whatever corner of the system it is if the lib met some requirements that should be well expressed in meson.build (checking version, headers, symbols).

I know, and that's okay. But the thing here is: we want to simulate a system where you can just install your VS, clone EFL, say "build it!" and it builds.

"But what about the dependencies?", a wise reader could ask. The answer, my dear reader, is: "the system should have them installed", but that's a cyclic question, so we need a more useful one. How about: "Do we know which ones are they?". That's the underlying question, which then, leads us to: "Are we missing a dependency that's not native in that given system? How do know?". All that applied to our current situation on our way to master the EFL (yes, pun intended) in Windows leads us to this short and brief...

**TL;DR:** We want to track which dependencies are or not native and not get surprised when someone would get stuck because of a dependency we didn't realized that exists (imagine pthreads coming from _somewhere unknown_ - parts would compile in a way, parts in another, and we wouldn't have how to track them easily). Strawberry-perl is just **a tool we used to compile OpenSSL**, which itself embeds some unix-default libraries which may mislead us, and not a dependency or whatever. Someone may use it to get the dependencies? Yes, but all our build starts from the principle of: "we're on a clean build and we got OpenSSL through magic".

> Are we aiming to manage the distribution of every dependency for native windows?

joaoantoniocardoso:
@JPTIZ thanks for such great explanation, now I think I understand the point :+1:

PR #191, opened 2020-06-29
Eio: Correctly use simple quotes on eio build definition

Coquinho:
`-DEIO_BUILD` was surrounded by ´\` instead of `''`.


Coquinho:
There are other things breaking Linux builds beside this.
Probably because now, there is no `EWAPI` defined when compiling `eo`

Coquinho:
Sooo, about the remaining error, we can just define `EWAPI` on eo for now....but it yields some errors about [eolian](https://travis-ci.com/github/expertisesolutions/efl/builds/173483175), so if we approve #171 it *should* resolve both errors. I will do some tests at `devs/coquinho/break-in-peaces/eo+eolian+efl`

PR #192, opened 2020-06-30
Implementing missing usleep()

joaoantoniocardoso:


src/lib/evil/evil_unistd.c
```diff
@@ -73,6 +73,20 @@ evil_time_get(void)
    return (double)_evil_time_second + (double)(count.QuadPart - _evil_time_count)/ (double)_evil_time_freq;
 }
 
+EVIL_API void
```
Coquinho, line 76: Maybe make it only for `_MSC_VER`? As it wasn't needed by msys.

src/lib/evil/evil_unistd.c
```diff
@@ -73,6 +73,20 @@ evil_time_get(void)
...
+usleep(__int64 usec)
```
walac, line 77: Why not the standard `uint64_t` value?

PR #194, opened 2020-06-30
Missing O_ACCMODE mask

joaoantoniocardoso:


src/lib/evil/evil_fcntl.h
```diff
@@ -3,9 +3,17 @@
...
-# include <sys/types.h>
+#include <sys/types.h>
 
 
+/**
+ * @def O_ACCMODE
+ * O_ACCMODE is an AND mask to extract file access modes.
+ */
+#ifdef _MSV_VER
+# define O_ACCMODE 3
```
JPTIZ, line 14: Commenting just in case in the future this leads to an issue: [here](https://github.com/pilight/windows-cli/blob/master/include/fcntl.h#L28) they used `O_RDONLY | O_WRONLY | O_RDWR` (which in practice will lead to the same result if their values is, respectively, 0, 1 and 2.

PR #197, opened 2020-06-30
Missing stdint and winsock2 includes

joaoantoniocardoso:


src/lib/evil/evil_time.h
```diff
@@ -4,7 +4,15 @@
...
 
 #include <time.h>
-
+#include <stdint.h>
+
+#ifdef _WIN32
+# ifndef WIN32_LEAN_AND_MEAN
+#  define WIN32_LEAN_AND_MEAN
+# endif
+#include "winsock2.h" // strcut timeval
```
JPTIZ, line 13: The famous "string cut".

(Typo)

joaoantoniocardoso, line 13:
```suggestion
#include "winsock2.h" // struct timeval
```

joaoantoniocardoso:
There are other problems related to winsock, I will fix it and reopen.

joaoantoniocardoso:
I think it is way more safe to include `evil_private.h` (that privately includes `evil_windows.h`) instead of directly including `windows.h` with `WIN32_LEAN_AND_MEAN` guards.

Since most of `WIN32_LEAN_AND_MEAN` was unnecessary I removed almost all of it.

JPTIZ:
> I think it is way more safe to include `evil_private.h` (that privately includes `evil_windows.h`) instead of directly including `windows.h` with `WIN32_LEAN_AND_MEAN` guards.
> 
> Since most of `WIN32_LEAN_AND_MEAN` was unnecessary I removed almost all of it.

And it is. I actually prefer it more, now that you mention...hm!

src/tests/evil/evil_test_unistd.c
```diff
@@ -25,9 +25,7 @@
 #include <io.h>
 #include <fcntl.h>
 
-# define WIN32_LEAN_AND_MEAN
-# include <winsock2.h>
-# undef WIN32_LEAN_AND_MEAN
+#include <winsock2.h>
```
JPTIZ, line 28: This one *must* be surrounded, tho. Or else it'll include `winsock.h` because of *reasons*.

joaoantoniocardoso: Are you sure? I did a bunch of tests here linking with evil, eina, eo, eolian, efl and ecore enabled and didn't noticed anything weird

JPTIZ: Well I guess the CI says I'm wrong
image

Coquinho: I think it [may include](https://gist.github.com/piscisaureus/906386#file-winsock2-h-L41) windows.h though

src/lib/evil/evil_private.h
```diff
@@ -56,9 +56,6 @@ extern "C" {
 #define sigsetjmp(Env, Save) setjmp(Env)
 
 #ifdef _MSC_VER
-#ifndef WIN32_LEAN_AND_MEAN
-# define WIN32_LEAN_AND_MEAN
```
Coquinho, line 60: Why removing this?

joaoantoniocardoso: It is not necessary as it simply prevents nothing: some file outside of evil could be including `windows.h` without including `evil_private.h` therefore not using this definition. Now those that are also including <evil_private.h>, that is correctly using `WIN32_LEAN_AND_MEAN` to prevent <windows.h> from including `<winsock.h>` instead of `<winsock2.h>`.

I think that if `WIN32_LEAN_AND_MEAN` should be globally defined it should be defined in `meson.build`.

joaoantoniocardoso:
> By changing `windowsn.h` and ´winsock.h`to`evil_private` aren't we including a lot of things that we don't need?

Yes, a viable option to that could be to declare `evil_windows.h` as a public header and include it.

src/lib/evil/evil_private.h
Coquinho: okay, as [basetsd;h](https://github.com/tpn/winsdk-10/blob/master/Include/10.0.10240.0/shared/basetsd.h) doesn't include any other windows header it should be ok.

Coquinho:
> > By changing `windowsn.h` and ´winsock.h`to`evil_private` aren't we including a lot of things that we don't need?
> 
> Yes, a viable option to deal with that could be to declare `evil_windows.h` as a public header and include it from outside.. 

Is it a problem to explicitly include windows headers? I mean, if at some point we need something from `winsock2.h` why should we include the whole `evil`? Why no including just what is needed?

>Or maybe just globally define `WIN32_LEAN_AND_MEAN` in `meson.build` and directly include `<windows.h>`.

I think globally could be a bad idea, we don't want everything that `windows.h/winsock2.h` includes but who knows about the user? So I think we shouldn't make it global.
Unless we can disable it at the end of compilation or something like that, so the symbol won't be defined for the user, is it possible?

joaoantoniocardoso:
> Is it a problem to explicitly include windows headers? I mean, if at some point we need something from winsock2.h why should we include the whole evil? Why no including just what is needed?

It is a problem to include `<windows.h>` without `WIN32_LEAN_AND_MEAN` defined if we are using `winsock2.h` and have no guarantee about the sequence of the includes.

> Unless we can disable it at the end of compilation or something like that, so the symbol won't be defined for the user, is it possible?

I think It could be done using `add_project_arguments` ([read here](https://mesonbuild.com/Adding-arguments.html#project-arguments))

> I think it may include windows.h though

Considering that the user may want to use `winsock.h` and not `winsock2.h` our current approach may leads to the same incompatibility caused [here](https://github.com/expertisesolutions/efl/pull/197/files#diff-73e6e76b85b517e802d375f88375e1aeL30).

src/tests/evil/evil_test_unistd.c
JPTIZ: It _did_ compile here, tho. I'm impressed.

JPTIZ: But, well, see the note: https://docs.microsoft.com/en-us/windows/win32/winsock/creating-a-basic-winsock-application

I have no idea why it works here

Coquinho:
> > Is it a problem to explicitly include windows headers? I mean, if at some point we need something from winsock2.h why should we include the whole evil? Why no including just what is needed?
> 
> It is a problem to include `<windows.h>` without `WIN32_LEAN_AND_MEAN` defined if we are using `winsock2.h` and have no guarantee about the sequence of the includes.
> 
> > Unless we can disable it at the end of compilation or something like that, so the symbol won't be defined for the user, is it possible?
> 
> I think It could be done using `add_project_arguments` ([read here](https://mesonbuild.com/Adding-arguments.html#project-arguments))
> 
> > I think it may include windows.h though
> 
> Considering that the user may want to use `winsock.h` and not `winsock2.h` our current approach may leads to the same incompatibility caused [here](https://github.com/expertisesolutions/efl/pull/197/files#diff-73e6e76b85b517e802d375f88375e1aeL30).

wait...so with `WIN_#@_LEAN_AND_MEAN` it doesn't work, but without it does? Oo

JPTIZ:
> > > By changing `windowsn.h` and ´winsock.h`to`evil_private` aren't we including a lot of things that we don't need?
> > 
> > 
> > Yes, a viable option to deal with that could be to declare `evil_windows.h` as a public header and include it from outside..
> 
> Is it a problem to explicitly include windows headers? I mean, if at some point we need something from `winsock2.h` why should we include the whole `evil`? Why no including just what is needed?

I don't think it is nice to insert platform-specific headers explicitly when we have platform-independent libraries, to be honest. Evil is there to deal with windows, even if it includes other things: why not, since we need general windows things in first place? It is there for it.

If the include is for specific things, tho, I don't mind including only that specific thing (i.e. `#include <winsock.h>`), although I would still prefer "the platform resolver library".

> 
> > Or maybe just globally define `WIN32_LEAN_AND_MEAN` in `meson.build` and directly include `<windows.h>`.
> 
> I think globally could be a bad idea, we don't want everything that `windows.h/winsock2.h` includes but who knows about the user? So I think we shouldn't make it global.

I also suggested in IRC about globally defining `WIN32_LEAN_AND_MEAN` and they also answered something in the line of "just where you need it, don't mess build system flags with this", and I agree.

> Unless we can disable it at the end of compilation or something like that, so the symbol won't be defined for the user, is it possible?

This is a case of leaving in `library`'s `c_args` but not on `declare_dependency`'s `compile_args`.

JPTIZ:
Also, in addition to including evil instead of windows: `windows.h` itself have _a lot_ of things already, what Evil adds is basically nothing compared to it.

src/generic/evas/common/timeout.c
```diff
@@ -1,12 +1,8 @@
 #ifdef _WIN32
 # include <stdio.h>
 
-# ifndef WIN32_LEAN_AND_MEAN
-#  define WIN32_LEAN_AND_MEAN
-# endif
-# include <windows.h>
+# include <evil_private.h>
 # include <process.h>
```
Coquinho, line 5: As here we are including `process.h` shouldn't it be maintained?

src/tests/evil/evil_test_unistd.c
joaoantoniocardoso: > But, well, see the note: https://docs.microsoft.com/en-us/windows/win32/winsock/creating-a-basic-winsock-application
> 
> I have no idea why it works here

The note says:

> The Winsock2.h header file internally includes core elements from the Windows.h header file, so there is not usually an #include line for the Windows.h header file in Winsock applications. If an #include line is needed for the Windows.h header file, this should be preceded with the #define WIN32_LEAN_AND_MEAN macro. For historical reasons, the Windows.h header defaults to including the Winsock.h header file for Windows Sockets 1.1. The declarations in the Winsock.h header file will conflict with the declarations in the Winsock2.h header file required by Windows Sockets 2.0. The WIN32_LEAN_AND_MEAN macro prevents the Winsock.h from being included by the Windows.h header. An example illustrating this is shown below.

And for what I am understanding this is almost exactly what we are doing with this patch. Am I missing something?

JPTIZ: There was one time I tried to compile and yes, winsock2 tried to include winsock. But could've been a side-effect.

joaoantoniocardoso: There was a file including `winsock.h` instead of  `winsock2.h`, maybe..?

JPTIZ: Nah, it was the same point, same `#include`, switching between `winsock` and `winsock2` - one would **always** clash with the other (winsock.h included winsock2.h, winsock2.h included winsock.h).

src/lib/evil/evil_strings.h
```diff
@@ -2,16 +2,7 @@
 #define EVIL_STRINGS_H__
 
 #include <evil_api.h>
```
JPTIZ, line 4: You can remove this if you're including `evil_api.h`.

src/tests/evil/evil_test_unistd.c
JPTIZ: Okay, `winsock2.h` indeed _potentially_ includes `Windows.h`: https://github.com/tpn/winsdk-10/blob/9b69fd26ac0c7d0b83d378dba01080e93349c2ed/Include/10.0.16299.0/um/WinSock2.h#L53

joaoantoniocardoso: yeah, it will include `windows.h` if it is not already included, as can be seen [here](https://github.com/tpn/winsdk-10/blob/9b69fd26ac0c7d0b83d378dba01080e93349c2ed/Include/10.0.10240.0/um/Windows.h#L25). Also if `windows.h` is included by `winsock2.h` then it will not try include `winsock.h` because `winsock2.h` prevents it [here](https://github.com/tpn/winsdk-10/blob/9b69fd26ac0c7d0b83d378dba01080e93349c2ed/Include/10.0.10240.0/um/WinSock2.h#L17)). Furthermore we can see [here](https://github.com/tpn/winsdk-10/blob/9b69fd26ac0c7d0b83d378dba01080e93349c2ed/Include/10.0.10240.0/um/Windows.h#L186) what is the impact of using `WIN32_LEAN_AND_MEAN` around `windows.h`.

src/lib/efreet/efreet_trash.c
```diff
@@ -30,11 +26,6 @@ static int _efreet_trash_log_dom = -1;
 static unsigned int _efreet_trash_init_count = 0;
 static const char *efreet_trash_dir = NULL;
 
-#ifdef _WIN32
-# define getuid() GetCurrentProcessId()
-#endif
-
-
```
Coquinho, line 37: Why removing this?

joaoantoniocardoso: [line 13](https://github.com/expertisesolutions/efl/pull/197/files/9c6caa5efd1855d134b2e3857f755f0a6abcf210#diff-f9abea168c370e10cbeaabb205fb0ccbR13) already defines it.

src/lib/eina/eina_debug.c
```diff
@@ -52,7 +52,7 @@
 
 #ifdef _WIN32
 # include <evil_private.h> /* fcntl */
-# include <winsock2.h> /* fcntl */
```
JPTIZ, line 55: lol

src/lib/evil/evil_time.h
```diff
@@ -3,8 +3,11 @@
 
 #include <evil_api.h>
 
+#include <stdint.h>
 #include <time.h>
+#include <stdint.h>
```
JPTIZ, line 8:
```suggestion
#include <stdint.h>
#include <time.h>
#include <stdint.h>
#include <stdint.h>
```

Y'know, just to make sure.

src/lib/efreet/efreet_trash.c
Coquinho: Why not removing the one from line 13? I think it make more sense as in <headers> <functions> separation.

PR #198, opened 2020-06-30
PACKAGE_SRC_DIR was using '\\' instead of '/'

joaoantoniocardoso:


Coquinho:
I don't understand much about what is happening here, I will wait for other reviewers.

JPTIZ:
To be honest, I've been thinking: should we even be using `\`? Windows apparently do understand `/` pretty ok, for example:

```console
tiz@WIN-GB1SN3JBMSR C:\Users\Tiz>cd source/repos 

tiz@WIN-GB1SN3JBMSR C:\Users\Tiz\source\repos>
```

JPTIZ:
Also we shouldn't be using this join/split. Meson has a builtin-function `join_paths` (or better yet: `/` operator) that works as a platform-independent path splitting, see this (the `"@0@"` is just because of quoting):

```meson
config_data = configuration_data({
    'PKG_BUILD_DIR' : '"@0@"'.format('foo' / 'bar'),
  })

config_h = configure_file(input : 'config.h.in',
                          output : 'config.h',
                          configuration : config_data)
```

With this:

```c
    printf("PKG_BUILD_DIR: %s\n", PKG_BUILD_DIR);
```

Prints this in both Linux and Windows:

```
PKG_BUILD_DIR: foo/bar
```

joaoantoniocardoso:
@JPTIZ the problem here is that the meson takes the the system path with `\` and not with `/` or `\\` on windows and then it is passed to C compiler using a escape sequence `\` as path separator...

JPTIZ:
> edit: I noticed that I confused my thoughts... I rewritten below:
> 
> @JPTIZ the problem here is that the meson takes the the system path with `\\` and not with `/` on windows and then it is passed to C compiler as a escape sequence `\`. On linux it is not a problem because meson takes the system path with '/'.
> 
> Try one simple mod in your example to understand what I mean.
> 
> ```diff
> ---    'PKG_BUILD_DIR' : '"@0@"'.format('foo' / 'bar'),
> +++    'PKG_BUILD_DIR' : '"@0@"'.format(meson.source_root()),
> ```
> 
> the result is: `#define PKG_BUILD_DIR "C:\Users\joaoantoniocardoso\Desktop\meson_path_test"`
> 
> but if you change it to:
> 
> ```diff
> ---    'PKG_BUILD_DIR' : '"@0@"'.format('foo' / 'bar'),
> +++    'PKG_BUILD_DIR' : '"@0@"'.format('/'.join(meson.source_root().split('\\'))),
> ```
> 
> the result now is: `#define PKG_BUILD_DIR "C:/Users/joaoantoniocardoso/Desktop/meson_path_test"`

Hmmm, in that case maybe we should try with `c_data.set_quoted` to see if it escapes.

joaoantoniocardoso:
> > edit: I noticed that I confused my thoughts... I rewritten below:
> > @JPTIZ the problem here is that the meson takes the the system path with `\\` and not with `/` on windows and then it is passed to C compiler as a escape sequence `\`. On linux it is not a problem because meson takes the system path with '/'.
> > Try one simple mod in your example to understand what I mean.
> > ```diff
> > ---    'PKG_BUILD_DIR' : '"@0@"'.format('foo' / 'bar'),
> > +++    'PKG_BUILD_DIR' : '"@0@"'.format(meson.source_root()),
> > ```
> > 
> > 
> > the result is: `#define PKG_BUILD_DIR "C:\Users\joaoantoniocardoso\Desktop\meson_path_test"`
> > but if you change it to:
> > ```diff
> > ---    'PKG_BUILD_DIR' : '"@0@"'.format('foo' / 'bar'),
> > +++    'PKG_BUILD_DIR' : '"@0@"'.format('/'.join(meson.source_root().split('\\'))),
> > ```
> > 
> > 
> > the result now is: `#define PKG_BUILD_DIR "C:/Users/joaoantoniocardoso/Desktop/meson_path_test"`
> 
> Hmmm, in that case maybe we should try with `c_data.set_quoted` to see if it escapes.

Yeah, it could work but I think that the original meson.build [already uses quoted](https://github.com/expertisesolutions/efl/pull/198/files#diff-969b60ad3d206fd45c208e266ccfed38L251).

JPTIZ:
> > > edit: I noticed that I confused my thoughts... I rewritten below:
> > > @JPTIZ the problem here is that the meson takes the the system path with `\\` and not with `/` on windows and then it is passed to C compiler as a escape sequence `\`. On linux it is not a problem because meson takes the system path with '/'.
> > > Try one simple mod in your example to understand what I mean.
> > > ```diff
> > > ---    'PKG_BUILD_DIR' : '"@0@"'.format('foo' / 'bar'),
> > > +++    'PKG_BUILD_DIR' : '"@0@"'.format(meson.source_root()),
> > > ```
> > > 
> > > 
> > > the result is: `#define PKG_BUILD_DIR "C:\Users\joaoantoniocardoso\Desktop\meson_path_test"`
> > > but if you change it to:
> > > ```diff
> > > ---    'PKG_BUILD_DIR' : '"@0@"'.format('foo' / 'bar'),
> > > +++    'PKG_BUILD_DIR' : '"@0@"'.format('/'.join(meson.source_root().split('\\'))),
> > > ```
> > > 
> > > 
> > > the result now is: `#define PKG_BUILD_DIR "C:/Users/joaoantoniocardoso/Desktop/meson_path_test"`
> > 
> > 
> > Hmmm, in that case maybe we should try with `c_data.set_quoted` to see if it escapes.
> 
> Yeah, it could work but I think that the original meson.build [already uses quoted](https://github.com/expertisesolutions/efl/pull/198/files#diff-969b60ad3d206fd45c208e266ccfed38L251).

Ok, this one now will be the "pffff ok" workaround of the week. Before anything, first see [this doc about strings](https://mesonbuild.com/Syntax.html#string-path-building), specifically the 3rd and 4th examples. Now look at this:

```meson
    'PKG_BUILD_DIR' : '"@0@"'.format(meson.source_root() / ''),
```

Guess what: this will generate...

```c
#define PKG_BUILD_DIR "C:/Users/Tiz/source/repos/mres/meson-paths/"
```

:D

JPTIZ:
Also, since `set_quoted` is already in use, you might want to return to that simple `set_quoted` with just this `/ ''` tweak.

joaoantoniocardoso:
@JPTIZ despite of adding a trailing `/` (that I don't think would be a problem), your suggestion is way more elegant, could you take a look again?

joaoantoniocardoso:
> Well, if `PACKAGE_SRC_DIR` is not used at all in windows...:D

Oh wait, wait.. `PACKAGE_SRC_DIR` is used by **edge** and **buildsystem** and because it was defined twice (first in `meson.build` and then in `src/static_libs/buildsystem/meson.build`) I removed from windows. _Maybe_ the same error will show up on build Linux too...

JPTIZ:
> > Well, if `PACKAGE_SRC_DIR` is not used at all in windows...:D
> 
> Oh wait, wait.. `PACKAGE_SRC_DIR` is used by **edge** and **buildsystem** and because it was defined twice (first in `meson.build` and then in `src/static_libs/buildsystem/meson.build`) I removed from windows. _Maybe_ the same error will show up on build Linux too...

Hm, mind opening an issue for this?

PR #200, opened 2020-06-30
[Ecore_con] Correctly define API

Coquinho:
Now there should be no `ECORE_CON_API extern ECORE_CON_API_WEAK`, only
`ECORE_CON_API ECORE_CON_API_WEAK extern`.

Places that `ECORE_CON_API_WEAK` is add were previously `EOAPI` that was declared
like `EWAPI`.

Some typos were corrected too, so some `ECORE_CON_WEAK_API` are now
`ECORE_CON_API_WEAK`

In summary I reverted the commit, made the folowing changes:
```
   extern EWAPI -> EWAPI extern
   EOAPI -> ECORE_CON_API ECORE_CON_API_WEAK
   EWAPI -> ECORE_CON_API ECORE_CON_API_WEAK
```
Reapplayed the commit.


Coquinho:
`evas` has problems too.

PR #204, opened 2020-07-01
[eet] Use EET_API instead of EAPI

Coquinho:
1. Create `lib_api.h`;
   `lib_api.h`;

definition of `LIB_API`.

### DIFF
--- a/src/lib/eet/eet_api.h
+++ b/src/lib/eet/eet_api.h
@@ -1,0 +1,34 @@
+#ifndef _EFL_CORE_API_H
+#define _EFL_CORE_API_H
+
+#ifdef EET_API
+#error EET_API should not be already defined
+#endif
+
+#ifdef _WIN32
+# ifndef EET_STATIC
+#  ifdef EET_BUILD
+#   define EET_API __declspec(dllexport)
+#  else
+#   define EET_API __declspec(dllimport)
+#  endif
+# else
+#  define EET_API
+# endif
+# define EET_API_WEAK
+#else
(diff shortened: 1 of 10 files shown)


src/lib/eet/eet_api.h
```diff
@@ -0,0 +1,34 @@
...
+# ifndef EFL_STATIC
+#  ifdef EET_BUILD
+#   define EET_API __declspec(dllexport)
+#  else
+#   define EET_API __declspec(dllimport)
+#  endif
+# else
+#  define EET_API
+# endif
+# define EET_API_WEAK
```
JPTIZ, line 18: `EET_WEAK_API` is not used. Should it be defined?

Coquinho: For now, the pattern I'm following is the one established by @felipealmeida, where it was defined.
But I agree with you on that, don't think we need to define it with it isn't used.

Coquinho: I would like to listen to other opinions here, guys.

joaoantoniocardoso: I think we should apply this patch as is;
if we are going to tailor each API to contain only the used symbols I think it would be more interest to do it when all this API idea gets more mature.

Coquinho: are you okay with that @JPTIZ ?

JPTIZ: I'm ok, yes.

PR #207, opened 2020-07-01
[ector] Correct typo at static definition

Coquinho:


joaoantoniocardoso:
Note that if only one symbol for the entire efl is what we aim - `EFL_STATIC` - I think we would need to make this patch to all API definitions merged last week.

Coquinho:
Yeah, it was a mistake.

PR #208, opened 2020-07-01
[ecore_input] Use ECORE_INPUT_API instead of EAPI

Coquinho:
1. Create `lib_api.h`;
   `lib_api.h`;

definition of `LIB_API`.

### DIFF
--- a/src/lib/ecore_input/meson.build
+++ b/src/lib/ecore_input/meson.build
@@ -20,7 +20,7 @@ endif
 
 ecore_input_lib = library('ecore_input',
     ecore_input_src, pub_eo_file_target,
-    c_args : package_c_args,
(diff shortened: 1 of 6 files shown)


src/lib/ecore_input/meson.build
```diff
@@ -20,7 +20,7 @@ endif
 
 ecore_input_lib = library('ecore_input',
     ecore_input_src, pub_eo_file_target,
-    c_args : package_c_args,
+    c_args : [package_c_args, '-DECORE_INPUT_BUILD'],
```
joaoantoniocardoso, line 23:
```suggestion
    c_args : [package_c_args + '-DECORE_INPUT_BUILD'],
```

PR #209, opened 2020-07-01
Implement eina_thread for native windows

walac:
Summary:
The implementation design respects the fact that Eina_Thread is an
uintptr_t. This means we need to allocate the thread struct in the heap
and return the pointer. This creates a complication for eina_thread_self,
as the returned Eina_Thread is not supposed to be explicitly freed (like
the posix implementation counterpart).

To solve this problem, we store the created thread structure in the
target thread TLS slot. For threads that were not created through eina
API, in eina_thread_self we allocate a new structure, push it to the TLS
slot and mark it to be freed on thread exit.

Notice we prefer _beginthreadex over CreateThread. _beginthreadex,
besides creating a new thread, ensures the C runtime will work on a
multithread environment.

Reviewers: jptiz, felipealmeida, vtorri

Subscribers: cedric, #reviewers, #committers, lucas

Tags: #efl

Differential Revision: https://phab.enlightenment.org/D12037


src/lib/eina/eina_sched.c
```diff
@@ -0,0 +1,84 @@
...
+
+#define RTNICENESS 1
+#define NICENESS 5
+
+void
+eina_sched_prio_drop(void)
+{
+   struct sched_param param;
+   int pol, ret;
+   pthread_t pthread_id;
```
Coquinho, line 42: Won't this be a problem?

walac: `eina_sched.c` is linux only, the implementation for windows is inside `eina_thread_win32.c`

Coquinho: Oh wait, every sched function on win32 is on eina_thread_win32? phab is ok with that? I think if we are doing alternative functions to posix implementation we should have `_posix/_win32` files

src/lib/eina/eina_thread_posix.c
```diff
@@ -0,0 +1,312 @@
...
+#include "eina_cpu.h"
+
+/* undefs EINA_ARG_NONULL() so NULL checks are not compiled out! */
+#include "eina_safety_checks.h"
+
+#include "eina_debug_private.h"
+
+#include <pthread.h>
+#include <errno.h>
+#ifndef _WIN32
```
Coquinho, line 39: Is it really needed?

walac: This is in case it runs in mingw

Coquinho: Won't mingw always use native threads?

src/lib/eina/eina_thread_posix.c
```diff
@@ -0,0 +1,312 @@
...
+   return r;
+}
+
+Eina_Thread
+eina_thread_self(void)
+{
+   return _eina_thread_self();
+}
+
+Eina_Bool
```
Coquinho, line 207: Why having an internal call definition and an external call on the same file? Why not only having `eina_thread_equal` that does what `_eina_thread_equal` does?

Coquinho: The same happens with join, equal and self

walac: This is from master. I didn't touch the code.

Coquinho: wtf, okay then

src/lib/eina/eina_thread_posix.c
walac: Now yes, but if someone wants to change the build files to test POSIX in a MinGW environment, it will work.

src/lib/eina/eina_sched.c
walac: I didn't touch sched in upstream yet, so this is the temporary solution to compile in native-windows without departing the implementation too much from phab.

src/lib/eina/eina_thread_posix.c
Coquinho: oh, nice, make sense

src/lib/eina/eina_sched.c
Coquinho: ohh understood, as it isn't the final solution it's okay then

src/lib/eina/eina_log.c
```diff
@@ -2189,7 +2187,7 @@ eina_log_print_cb_journald(const Eina_Log_Domain *d,
                                             "PRIORITY=%i", level,
                                             "MESSAGE=%s", message,
                                             "EFL_DOMAIN=%s", d->domain_str,
-                                            "THREAD=%lu", eina_thread_self_id(),
+                                            "THREAD=%lu", eina_thread_self(),
```
JPTIZ, line 2190: Does it mean we're back to consider `self` can return the ID/integer type?

JPTIZ:
Btw, the description is outdated both here and in phab (`sys_native_windows` doesn't exist anymore). Did something else special change since the patch was first submited?

src/lib/eina/eina_log.c
Coquinho: Now a thread is always an `intptr_t`

walac:
> Btw, the description is outdated both here and in phab (sys_native_windows doesn't exist anymore). Did something else special change since the patch was first submited?

Oh, thanks, I need to update the description, thanks.

src/lib/eina/eina_log.c
JPTIZ: Oh ok.

src/lib/eina/eina_thread_win32.c
```diff
@@ -0,0 +1,390 @@
...
+                   int affinity, Eina_Thread_Cb func, const void *data)
+{
+   Thread_t *thr = calloc(1, sizeof(Thread_t));
+   if (!thr)
+      return EINA_FALSE;
+
+   thr->data = (void *) data;
+   thr->fn = func;
+
+   thr->handle = (HANDLE) _beginthreadex(NULL, 0, thread_fn, thr, CREATE_SUSPENDED, &thr->id);
```
JPTIZ, line 131: We don't need any of the operations from `eina_internal_call` then? AFAIK Win32 threads aren't cancelable, so this part is fine I guess, but what about the cleanup_push/pop with debug operations?

Coquinho: Yeah, debug info is always good :D

walac: > but what about the cleanup_push/pop with debug operations?

@JPTIZ what do you mean?

JPTIZ: @walac in `eina_thread_posix.c` we always wrapped the given function with `_eina_internal_call`, which itself called those `EINA_THREAD_CLEANUP_PUSH`/`EINA_THREAD_CLEANUP_POP` with some debugging functions.

JPTIZ: Push with free function (no idea why): https://github.com/expertisesolutions/efl/pull/209/files/75acf308286175d146310f47dfe944a297e1ffb3#diff-034d507343e6df83b23e5ca8aaf7c37fR143

Push with debug function: https://github.com/expertisesolutions/efl/pull/209/files/75acf308286175d146310f47dfe944a297e1ffb3#diff-034d507343e6df83b23e5ca8aaf7c37fR193
And both Pops: https://github.com/expertisesolutions/efl/pull/209/files/75acf308286175d146310f47dfe944a297e1ffb3#diff-034d507343e6df83b23e5ca8aaf7c37fR195

JPTIZ: Oh, ok, the `free` is explained here: https://github.com/expertisesolutions/efl/pull/209/files/75acf308286175d146310f47dfe944a297e1ffb3#diff-034d507343e6df83b23e5ca8aaf7c37fR233

joaoantoniocardoso:
It fixes the problems with ecore but when compiling, the throws an error about `eo`. maybe this diff does the job:

```diff
diff --git a/src/lib/eo/eo_ptr_indirection.c b/src/lib/eo/eo_ptr_indirection.c
index 80622e789c..a342690b0c 100644
--- a/src/lib/eo/eo_ptr_indirection.c
+++ b/src/lib/eo/eo_ptr_indirection.c
@@ -34,7 +34,6 @@ _eo_obj_pointer_invalid(const Eo_Id obj_id,
                         int line)
 {
    Eina_Thread thread = eina_thread_self();
-   Eina_ThreadId thread_id = eina_thread_self_id();
    const char *tself = "main";
    const char *type = "object";
    const char *reason = "This ID has probably been deleted";
@@ -42,7 +41,7 @@ _eo_obj_pointer_invalid(const Eo_Id obj_id,
    if (obj_id & MASK_CLASS_TAG) type = "class";
    if (!eina_thread_equal(thread, _efl_object_main_thread))
      {
-        snprintf(tbuf, sizeof(tbuf), "%ld", (long) thread_id);
+        snprintf(tbuf, sizeof(tbuf), "%lu", (unsigned long)eina_thread_self());
         tself = tbuf;
      }
```

src/lib/eina/eina_thread_win32.c
walac: Ok, added!

Coquinho:
@walac what was the change?

walac:
@Coquinho addressed @joaoantoniocardoso comment and add debug like in `_eina_internal_call`

Coquinho:
> @Coquinho addressed @joaoantoniocardoso comment and add debug like in `_eina_internal_call`

nice, I will have a look

PR #212, opened 2020-07-01
[efreet] Use EFREET_API instead of EAPI

Coquinho:
1. Create `lib_api.h`;
   `lib_api.h`;

definition of `LIB_API`.

### DIFF
--- a/src/lib/efreet/Efreet_Mime.h
+++ b/src/lib/efreet/Efreet_Mime.h
@@ -16,6 +16,10 @@
 
 #include <efreet_api.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /**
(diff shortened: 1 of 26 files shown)


src/lib/efreet/Efreet_Mime.h
```diff
@@ -14,86 +14,57 @@
...
-#   define EAPI
-#  endif
-# else
-#  define EAPI
-# endif
-#endif
-
-#ifdef __cplusplus
-extern "C" {
-#endif
```
walac, line 45: Why the wrap around C++ has been removed?

Coquinho: wow, it is a mistake, sorry about that, and thanks for noticing.

PR #216, opened 2020-07-01
[ephysics] Use EPHYSICS_API instead of EAPI

Coquinho:
1. Create `lib_api.h`;
   `lib_api.h`;

definition of `LIB_API`.

### DIFF
--- a/src/lib/ephysics/ephysics_api.h
+++ b/src/lib/ephysics/ephysics_api.h
@@ -1,0 +1,34 @@
+#ifndef _EFL_EPHYSICS_API_H
+#define _EFL_EPHYSICS_API_H
+
+#ifdef EPHYSICS_API
+#error EPHYSICS_API should not be already defined
+#endif
+
+#ifdef _WIN32
+# ifndef EFL_STATIC
+#  ifdef EPHYSICS_BUILD
+#   define EPHYSICS_API __declspec(dllexport)
+#  else
+#   define EPHYSICS_API __declspec(dllimport)
+#  endif
+# else
+#  define EPHYSICS_API
+# endif
+# define EPHYSICS_API_WEAK
+#else
+# ifdef __GNUC__
+#  if __GNUC__ >= 4
+#   define EPHYSICS_API __attribute__ ((visibility("default")))
+#   define EPHYSICS_API_WEAK __attribute__ ((weak))
+#  else
+#   define EPHYSICS_API
+#   define EPHYSICS_API_WEAK
+#  endif
+# else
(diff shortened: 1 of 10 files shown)


src/lib/ephysics/ephysics_api.h
```diff
@@ -6,7 +6,7 @@
 #endif
 
 #ifdef _WIN32
-# ifndef EFL_STATIC
+# ifndef EPHYSICS_STATIC
```
JPTIZ, line 9: But isn't `EFL_STATIC` global?

JPTIZ: Or better: isn't global the idea of compiling statically? I mean...we either compile the whole EFL statically or dinamically.

Coquinho: Please take a look [here](https://expertisesolutions.slack.com/archives/C7NGCQY8M/p1593608488004500).
In my mind what you are saying makes sense, but it seems that it was intended from the beginning to be like this, of course, that this could be a mistake. 
So it really wouldn't make sense compiling only one lib as static right, @JPTIZ ?

JPTIZ: Oh ook ok. Actually all of this is not the idea.

So, first:
- We're **not** getting a static EFL in the future. This is already a possibility (all you have to do is _not_ define the `DLL_EXPORT` macro).
- The idea is actually the opposite: in the future, EFL will have **no** static support. Yes, EFL's aiming into removing static support once for all.

In the end, this ifndef shouldn't exist (or, in the best case, consider `DLL_EXPORT` instead)

joaoantoniocardoso: I would suggest to address an issue to work on this after we managed to get this working. I don't think this static thing is hurting for now...

JPTIZ: > I would suggest to address an issue to work on this after we managed to get this working. I don't think this static thing is hurting for now...

In this case, I would strongly suggest to change to `DLL_EXPORT` (and then approve + create issue) since this is how static building is being managed nowadays, even because dropping static support is something the upstream is wanting to do, so...why should we bother with what they want?

joaoantoniocardoso: Yeah, i understand your point, I just would fix this kind of change using automation and it is way easier when every file follows the same structure.

JPTIZ: Well, I won't end Coquinho's happiness here after so many patches. I'll leave that for later then.

PR #196, opened 2020-06-30
Adding missing ST*_FILENO definitions

joaoantoniocardoso:

PR #223, opened 2020-07-02
Typo in _fileno, missing stdio include

joaoantoniocardoso:
Fixing a mistake of #196.

### DIFF
--- a/src/lib/evil/evil_unistd.h
+++ b/src/lib/evil/evil_unistd.h
@@ -16,6 +16,7 @@
 
 #ifdef _WIN32
 
+#include <stdio.h>
 #include <io.h> // for read, write, access, close
 
 #define execvp _ucrt_execvp  // overriding execvp below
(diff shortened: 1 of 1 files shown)


JPTIZ:
> Fixing a mistake of #196.

What a terrible mistake. How did we let that pass?

src/lib/evil/evil_unistd.h
```diff
@@ -30,9 +30,10 @@ EVIL_API int ftruncate(int fd, off_t size);
 #define X_OK    0       /* execute permission, originally '1', just a bypass here*/
 #define F_OK    0       /* Test for existence.  */
 
-#define STDIN_FILENO __fileno(stdin)
-#define STDOUT_FILENO __fileno(stdout)
-#define STDERR_FILENO __fileno(stderr)
+#include <stdio.h>
```
Coquinho, line 33: maybe add it in the beginning of the file with other includes?

JPTIZ: Yes, you're right. Didn't see that it wasn't the beginning of the file.

joaoantoniocardoso: done

PR #225, opened 2020-07-02
extern char **_environ is already defined by ucrt.

joaoantoniocardoso:
ref: https://docs.microsoft.com/en-us/cpp/c-runtime-library/environ-wenviron?view=vs-2019


joaoantoniocardoso:
> Nice catch. But UCRT defines `_environ`, not `environ`. Maybe you would want to change it into a `#define` (and tbh this looks like should be in evil, or am I wrong?).

Now that you commented I noticed that I really need to take a better look. I'll reopen when its ready again.

Coquinho:
Assuming this.

joaoantoniocardoso:
I can't review because I started this PR, but this makes way more sense than the initial work.

PR #226, opened 2020-07-02
windows: Fix tls free callback

walac:
The callback is only supposed to be called on thread exit.

### DIFF
--- a/src/lib/eina/eina_inline_lock_win32.x
+++ b/src/lib/eina/eina_inline_lock_win32.x
@@ -446,7 +446,6 @@ static inline void
 _eina_tls_free(Eina_TLS key)
 {
    eina_lock_take(&_eina_tls_map_lock);
-   _eina_free_tls_value(&key, TlsGetValue(key));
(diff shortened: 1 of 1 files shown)


src/lib/eina/eina_inline_lock_win32.x
```diff
@@ -461,10 +460,8 @@ _eina_tls_get(Eina_TLS key)
 static inline Eina_Bool
 _eina_tls_set(Eina_TLS key, const void *data)
 {
-   void *p = TlsGetValue(key);
    const Eina_Bool ret = TlsSetValue(key, (void *) data) ? EINA_TRUE : EINA_FALSE;
    eina_lock_take(&_eina_tls_map_lock);
-   if (ret) _eina_free_tls_value(&key, p);
    eina_lock_release(&_eina_tls_map_lock);
```
Coquinho, line 465: Do we still need this lock/release than?

walac: You are right, fixed. Thanks.

PR #227, opened 2020-07-02
eina_fnmatch.c: Fix buffer overruns

walac:


src/lib/eina/eina_fnmatch.c
```diff
@@ -103,51 +103,51 @@ _wildcards_to_regex(const char *pattern, int flags)
                             if ( (j == 0) || (pattern[j - 1] == '/') )
                               {
                                  strcpy_s(reg_pattern + i,
-                                          sizeof(reg_pattern + i),
+                                          reg_length - i,
```
Coquinho, line 106: Why `-i`? Didn't got it.

JPTIZ: The second argument is the length. We're starting from `i` chars after the start of `reg_pattern`, then our length will be the pattern's length minus where we're starting. Like with `"hello, world!"` (length 13): if we start from `", world"` (index 5, length 8), we want to copy 8 chars, which is 13 - 5 (length - new starting index).

JPTIZ: Also, just for reference, @walac: will the resulting string still be null-terminated?

walac: Yes, I replaced the alloc function by `calloc` we provides a zero filled buffer.

PR #229, opened 2020-07-02
[ECORE_EVAS] Use ECORE_EVAS_API instead of EAPI

Coquinho:
1. Create `lib_api.h`;
   `lib_api.h`;

definition of `LIB_API`.

### DIFF
--- a/src/lib/ecore_evas/ecore_evas_private.h
+++ b/src/lib/ecore_evas/ecore_evas_private.h
@@ -1,41 +1,17 @@
 #ifndef _ECORE_EVAS_PRIVATE_H
 #define _ECORE_EVAS_PRIVATE_H
 
-#ifdef EAPI
-# undef EAPI
-#endif
-
-#ifdef _WIN32
-# ifdef EFL_BUILD
-#  ifdef DLL_EXPORT
-#   define EAPI __declspec(dllexport)
-#  else
-#   define EAPI
-#  endif
-# else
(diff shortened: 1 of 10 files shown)


src/lib/ecore_evas/ecore_evas_private.h
```diff
@@ -442,30 +418,30 @@ struct _Ecore_Evas_Aux_Hint
...
-EAPI void _ecore_evas_register_animators(Ecore_Evas *ee);
-EAPI void _ecore_evas_free(Ecore_Evas *ee);
-EAPI void _ecore_evas_idle_timeout_update(Ecore_Evas *ee);
-EAPI void _ecore_evas_mouse_move_process(Ecore_Evas *ee, int x, int y, unsigned int timestamp);
-EAPI void _ecore_evas_mouse_device_move_process(Ecore_Evas *ee, Efl_Input_Device *pointer,
+ECORE_EVAS_API void _ecore_evas_ref(Ecore_Evas *ee);
+ECORE_EVAS_API void _ecore_evas_unref(Ecore_Evas *ee);
+ECORE_EVAS_API int ecore_evas_buffer_render(Ecore_Evas *ee);
+
+ECORE_EVAS_API void _ecore_evas_fps_debug_init(void);
```
joaoantoniocardoso, line 425: before someone asks, I checked this some days ago and I found that all this `_functions` seems to be intern but actually needs to be exported.

PR #235, opened 2020-07-02
build: Disable pkgconfig operations for MSVC (Fixes #150)

### LINKED ISSUE #150: Change pkgconfig into disabler-object on Windows
As I mentioned in #149, instead of making tons of `if not sys_windows` when pkgconfig is used, we could simply change into a [`disabler()` object](https://mesonbuild.com/Reference-manual.html#disabler-object).

JPTIZ:
Since MSVC builds don't know how to deal with Pkg-config, this commit changes them into a [disabler-object](https://mesonbuild.com/Reference-manual.html#disabler-object), which makes all operations using the variable `pkgconfig` to be disabled (i.e. ignored).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -8,10 +8,6 @@ if host_machine.system() == 'darwin'
   add_languages('objc')
 endif
 
-if host_machine.system() != 'windows'
-  pkgconfig = import('pkgconfig')
-endif
-
 test_env = environment()
 test_env.set('EFL_RUN_IN_TREE', '1')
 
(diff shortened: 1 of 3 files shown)


Coquinho:
Didn't break anything, should I test it in any other way?

meson.build
```diff
@@ -64,6 +60,12 @@ sys_bsd = bsd.contains(host_machine.system())
 sys_windows = windows.contains(host_machine.system())
 sys_osx = osx.contains(host_machine.system())
 
+if sys_windows and cc.get_id() == 'msvc'
```
walac, line 63: should test for `clang-cl` too

JPTIZ: Done o/

PR #236, opened 2020-07-02
Use diferente names for EO events

walac:


JPTIZ:
What is this fixing/improving exactly? Debugging?

Also, if the name itself is important and since we're already making changes, maybe changing to "tester", "tester_subscribe" and "tester_clamp_test"?

Coquinho:
> What is this fixing/improving exactly? Debugging?
> 
> Also, if the name itself is important and since we're already making changes, maybe changing to "tester", "tester_subscribe" and "tester_clamp_test"?

It seems like a good idea!

walac:
> It seems like a good idea!

Done

PR #239, opened 2020-07-02
[ecore_wayland] Use ECORE_WAYLAND_API instead of EAPI

Coquinho:
1. Create `lib_api.h`;
   `lib_api.h`;

definition of `LIB_API`.


walac:
Do we care about wayland?

Coquinho:
> Do we care about wayland?

No, we don't, but at least for me don't make sense to let some libraries in a pattern and others in another one.

PR #241, opened 2020-07-02
[Ecore] Fixing a dll export

joaoantoniocardoso:
I think this functions should be intern and therefore don't need to be exported. If it is not the case, its declaration in .h is missing the `ECORE_EAPI` symbol.


Coquinho:
Which error does it yields?

joaoantoniocardoso:
> Which error does it yields?

It was a warning about the linker failing to export that definition.

Coquinho:
I will wait for someone with any more knowledge in linking to say something before I approve. For me that change makes sense.

PR #246, opened 2020-07-03
Fixing undefined useconds_t

joaoantoniocardoso:
* now `usleep()` is implemented in evil (PR #192), then we don't need to keep the `Sleep()` specifically for windows, and then we don't need to include `windows.h` here.

 * stdint.h include removed because it was only needed because of uint64_t.

 * from [IEEE](https://pubs.opengroup.org/onlinepubs/007904975/basedefs/sys/types.h.html): _The type useconds_t shall be an unsigned integer type capable of storing values at least in the range [0, 1000000]. The type suseconds_t shall be a signed integer type capable of storing values at least in the range [-1, 1000000]._, -> then it is here defined as `unsigned int`.

### DIFF
--- a/src/lib/evas/common/evas_image_scalecache.c
+++ b/src/lib/evas/common/evas_image_scalecache.c
@@ -2,7 +2,7 @@
 # include "config.h"
 #endif
 
-#ifdef _MSC_VER
+#ifdef _WIN32
 # include <evil_private.h>
 #endif
 
--- a/src/lib/evas/common/evas_image_scalecache.c
+++ b/src/lib/evas/common/evas_image_scalecache.c
@@ -2,12 +2,8 @@
 # include "config.h"
 #endif
 
-#ifdef _WIN32
-# ifndef WIN32_LEAN_AND_MEAN
-#  define WIN32_LEAN_AND_MEAN
-# endif
-# include <windows.h>
-# undef WIN32_LEAN_AND_MEAN
+#ifdef _MSC_VER
+# include <evil_private.h>
 #endif
 
 #include <assert.h>
@@ -541,11 +537,7 @@ evas_common_rgba_image_scalecache_prepare(Image_Entry *ie, RGBA_Image *dst EINA_
         
         while (slpt < 500000)
           {
-#ifdef _WIN32
-             Sleep(slp / 1000);
-#else
              usleep(slp);
-#endif
              slpt += slp;
              slp++;
              ret = SLKT(im->cache.lock);
--- a/src/lib/evil/evil_unistd.h
+++ b/src/lib/evil/evil_unistd.h
@@ -62,8 +62,9 @@ EVIL_API int ftruncate(int fd, off_t size);
 EVIL_API double evil_time_get(void);
 
 #ifdef _MSC_VER
-# include <stdint.h>
-EVIL_API void usleep(uint64_t usec);
+typedef unsigned long useconds_t;
+
+EVIL_API void usleep(useconds_t usec);
 #endif
 
 /*
--- a/src/lib/evil/evil_unistd.h
+++ b/src/lib/evil/evil_unistd.h
@@ -62,7 +62,7 @@ EVIL_API int ftruncate(int fd, off_t size);
 EVIL_API double evil_time_get(void);
 
 #ifdef _MSC_VER
-typedef unsigned long useconds_t;
+typedef unsigned int useconds_t;
 
 EVIL_API void usleep(useconds_t usec);
 #endif
--- a/src/lib/evil/evil_unistd.c
+++ b/src/lib/evil/evil_unistd.c
@@ -95,7 +95,7 @@ evil_time_get(void)
 
 #ifdef _MSC_VER
 EVIL_API void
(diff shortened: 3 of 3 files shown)


src/lib/evas/common/evas_image_scalecache.c
```diff
@@ -537,15 +533,11 @@ evas_common_rgba_image_scalecache_prepare(Image_Entry *ie, RGBA_Image *dst EINA_
    ret = SLKT(im->cache.lock);
    if (ret == EINA_FALSE) /* can't get image lock */
      {
-        useconds_t slp = 1, slpt = 0;
```
JPTIZ, line 540: I think instead we could create a `typedef` for `useconds_t` in Evil, what do you think? This way we ensure we're still using what posix expected :D

joaoantoniocardoso: Yeah, I agree

src/lib/evil/evil_unistd.h
```diff
@@ -62,8 +62,9 @@ EVIL_API int ftruncate(int fd, off_t size);
 EVIL_API double evil_time_get(void);
 
 #ifdef _MSC_VER
-# include <stdint.h>
```
Coquinho, line 65: we don't need stdint anymore?

joaoantoniocardoso: > * stdint.h include removed because it was only needed because of uint64_t

JPTIZ: Why not using `uint64_t` again (instead of `unsigned long`)?

Coquinho: > > * stdint.h include removed because it was only needed because of uint64_t

oh, please add this to the PR description as well.

joaoantoniocardoso: > Why not using `uint64_t` again (instead of `unsigned long`)?

I don't understand the point of including another header to define a type that we know that is a fixed-width integral. Actually shouldn't be `unsigned long long`?

JPTIZ: Mostly because `uint64_t` is the way to go in these cases, since:
- It is platform independent (this is the main point);
- It is what we should use if we want an integral data type with specific length (which enforces the previous point).

But seeing [mingw's sys/types.h](https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/crt/sys/types.h#l84), it defines `useconds_t` to `unsigned int`. Maybe that should be its final type? What's POSIX's one?

joaoantoniocardoso: > The type useconds_t shall be an unsigned integer type capable of storing values at least in the range [0, 1000000]. The type suseconds_t shall be a signed integer type capable of storing values at least in the range [-1, 1000000].

ref: https://pubs.opengroup.org/onlinepubs/007904975/basedefs/sys/types.h.html

JPTIZ: Hm, in that case, `unsigned int` (which is guaranteed by C and C++ specifications to have >= 4 byte size) is more than enough, since 1000000 fits in 20 bits.

src/lib/evil/evil_unistd.h
```diff
@@ -62,8 +62,9 @@ EVIL_API int ftruncate(int fd, off_t size);
...
-EVIL_API void usleep(uint64_t usec);
+typedef unsigned long useconds_t;
```
joaoantoniocardoso, line 65:
```suggestion
typedef unsigned int useconds_t;
```

JPTIZ: Yep.

src/lib/evil/evil_unistd.h
joaoantoniocardoso: done

src/lib/evas/common/evas_image_scalecache.c
```diff
@@ -2,12 +2,8 @@
 # include "config.h"
 #endif
 
-#ifdef _WIN32
-# ifndef WIN32_LEAN_AND_MEAN
-#  define WIN32_LEAN_AND_MEAN
-# endif
-# include <windows.h>
-# undef WIN32_LEAN_AND_MEAN
+#ifdef _MSC_VER
```
Coquinho, line 5: So there is no other thing that `cygwin/msys` were using here?

joaoantoniocardoso: I just checked the blame and prior to `windows.h` it was using `Evil.h`.

Coquinho: ohh, so shouldn't this be like?
```
#ifdef _WIN32
 #include <evil_private.h>
Endif
```

joaoantoniocardoso: Yes it should. Thanks

joaoantoniocardoso, line 5:
```suggestion
#ifdef _WIN32
```

PR #250, opened 2020-07-03
[Emile] [Redone] Building with tests

joaoantoniocardoso:
Turns out that emile doesn't need ecore, so I redone the work of PR #120 here, sending it directly to windows-native.

Test:
```
23/23 emile-suite               OK             0.10s
```

PR #252, opened 2020-07-03
build: Fix libjpeg trying to link with freetype

JPTIZ:


Coquinho:
Is this PR ready for review? Its kind confuse when the title says "try"

JPTIZ:
> Is this PR ready for review? Its kind confuse when the title says "try"

Yes, it is. The meaning is "libjpeg currently tries to link with freetype. This patches fixes it by making it link to the right one"

Coquinho:
oh okay

Coquinho:
How can we see if ilbjpeg is being imported correctly?

JPTIZ:
> How can we see if ilbjpeg is being imported correctly?

I think this will have to be done locally with `-Dnuget-download=true`. Also, shouldn't we have a CI target with this? (And another without)

Coquinho:
> > How can we see if ilbjpeg is being imported correctly?
> 
> I think this will have to be done locally with `-Dnuget-download=true`. Also, shouldn't we have a CI target with this? (And another without)

Don't know, how could this work?

JPTIZ:
> > > How can we see if ilbjpeg is being imported correctly?
> > 
> > 
> > I think this will have to be done locally with `-Dnuget-download=true`. Also, shouldn't we have a CI target with this? (And another without)
> 
> Don't know, how could this work?

Just make a clone of the current Windows pipeline, but adding the flag.

joaoantoniocardoso:
@JPTIZ  Did you noticed that this is being fixed [here](https://github.com/expertisesolutions/efl/pull/250/files) too?

JPTIZ:
> @JPTIZ Did you noticed that this is being fixed [here](https://github.com/expertisesolutions/efl/pull/250/files) too?

_Shhhhh!_

Coquinho:
So this will be closed?

Issue #248, opened 2020-07-03
[Linux] [To investigate] Maybe PACKAGE_SRC_DIR is defined twice


joaoantoniocardoso:
From [this discussion](https://github.com/expertisesolutions/efl/pull/198#issuecomment-653623793), maybe `PACKAGE_SRC_DIR` is defined twice on linux too, being the first in `meson.build` and then in `src/static_libs/buildsystem/meson.build`.

joaoantoniocardoso:
Definitely it should be removed for linux too:
```
In file included from ../src/static_libs/buildsystem/buildsystem_meson.c:1:
./config.h:416: warning: "PACKAGE_SRC_DIR" redefined
  416 | #define PACKAGE_SRC_DIR "/home/joaoantoniocardoso/ExpertiseSolutions/efl2/"
```

PR #259, opened 2020-07-05
Removing redefinition of PACKAGE_SRC_DIR for linux

joaoantoniocardoso:
This should solve #248.


Coquinho:
So this revert halve of #198, it will still work as intended?

PR #260, opened 2020-07-05
Fixing ecore_x linkage to elementary examples

joaoantoniocardoso:
Yes, its odd but Elementary should be included first or Ecore_X won't link on linux.

### DIFF
--- a/src/examples/elementary/meson.build
+++ b/src/examples/elementary/meson.build
@@ -124,7 +124,7 @@ examples = [
 ]
 
 foreach example : examples
-  executable(example, example + '.c', dependencies: [elementary, ecore, eio, m])
+  executable(example, example + '.c', dependencies: [elementary, ecore, ecore_x, eio, m])
 endforeach
 if get_option('bindings').contains('cxx')
   cxx_examples = [
--- a/src/examples/elementary/win_example.c
+++ b/src/examples/elementary/win_example.c
@@ -1,11 +1,11 @@
 /*
  * gcc -o win_example win_example.c `pkg-config --cflags --libs elementary ecore-x`
  */
+#include <Elementary.h>
+
 #ifdef HAVE_ELEMENTARY_X
 # include <Ecore_X.h>
 #endif
-
-#include <Elementary.h>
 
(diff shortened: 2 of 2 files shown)


Coquinho:
If we are going to explicitly include ecore_x I would prefer #144, As it only adds ecore_x when needed.

joaoantoniocardoso:
> If we are going to explicitly include ecore_x I would prefer #144, As it only adds ecore_x when needed.

I completely forgot that PR and... Ok, I also prefer #144.

> Oh it is that thing again. The weird part: `ecore_x` is already a dependency of `elementary`...but somehow it just blows up for no reason.

welp, where is it declared?

Coquinho:
> > If we are going to explicitly include ecore_x I would prefer #144, As it only adds ecore_x when needed.
> 
> I completely forgot that PR and... Ok, I also prefer #144.
> 
> > Oh it is that thing again. The weird part: `ecore_x` is already a dependency of `elementary`...but somehow it just blows up for no reason.
> 
> welp, where is it declared?

It is declared at elementary's [meson.build](https://github.com/expertisesolutions/efl/blob/0132b75586fed26f80f9c469390565b797d46cc5/src/lib/elementary/meson.build#L990)

Coquinho:
> Oh it is that thing again. The weird part: `ecore_x` is already a dependency of `elementary`...but somehow it just blows up for no reason.

And the best part, somehow it doesn't blow at masters

Coquinho:
> > If we are going to explicitly include ecore_x I would prefer #144, As it only adds ecore_x when needed.
> 
> I completely forgot that PR and... Ok, I also prefer #144.

@joaoantoniocardoso so I will make It ready.

joaoantoniocardoso:
> > > If we are going to explicitly include ecore_x I would prefer #144, As it only adds ecore_x when needed.
> > 
> > 
> > I completely forgot that PR and... Ok, I also prefer #144.
> 
> @joaoantoniocardoso so I will make It ready.

Closing this one in favor of #144.

JPTIZ:
Suppressed by #144.

Issue #218, opened 2020-07-02
eolian_gen doesn't accept Windows absolute paths as the output's base name


cauebs:
From eolian_gen's usage string:
```
-g type       generate file of type "type"
-o name       specify the base name for output
-o type:name  specify a particular output filename
```
The base name is used in the path of each output file generated (the ones specified by `-g`), unless there is an explicit `-o type:name`. On Windows, absolute paths contain a colon (e.g.: `C:\...`) and that messes up the parsing, leading to a runtime error: `unknown type: C`.

cauebs:
Issue moved upstream: https://phab.enlightenment.org/T8767

PR #263, opened 2020-07-06
Clean up eolian_generation test to work around #218

cauebs:
Works around #218 so that eolian_generation tests can be ran on Windows.

Changes helper functions to pass around the filename already with the extension and use `-o type:filename` instead of `-o basename`.

### DIFF
--- a/src/tests/eolian/eolian_generation.c
+++ b/src/tests/eolian/eolian_generation.c
@@ -11,20 +11,17 @@
 #include "eolian_suite.h"
 
 static Eina_Bool
-_files_compare (const char *ref_filename, const char *tmp_filename, const char *ext)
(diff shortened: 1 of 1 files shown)


cauebs:
> Was this intended to make eolian tests pass?

No, it makes them run. Before, the call to `eolian_gen` would fail. A subsequent PR will address the failing tests.

cauebs:
With this last commit and #264, all eolian tests pass.

cauebs:
> Shouldn't we remove _after_, then?

Not necessarily. Files are created in a temp directory and they are deleted before each test to avoid false positives.

JPTIZ:
Suppressed by #144

JPTIZ:
WRONG TAB

PR #265, opened 2020-07-06
CI: Windows: Enable tests

Coquinho:
The idea is to  create another windows CI so we can see how a branch is
going about the unity tests.


.github/workflows/windows-test.yml
```diff
@@ -0,0 +1,93 @@
...
+      run: nmake
+      working-directory: .\openssl
+    - name: configure
+      run: .\configure.bat
+      env:
+        OPENSSL_DIR: ${{ github.workspace }}\openssl
+    - name: Build
+      run: .\build.bat
+    - name: Tests
+      run: .\test.bat
```
JPTIZ, line 93:
```suggestion
  tests:
    - name: Tests
      run: .\test.bat
```

JPTIZ:
Move changes to original `windows.yml` :)

JPTIZ:
Just for clarification: having an additional `windows-test.yml` adds another "machine" to the CI that'll also build the EFL, so we'll have:

- `windows.yml`: Runs `build` job.
- `windows-test.yml`: Runs `build` and then `tests` job.

Which means we'll be running the same `build` two times (and having to make sure that when one is changed, the other is also changed). That means that we're duplicating jobs here: if one `build` fails obviously the other `build` will fail too - it is *the same* procedure.

By just adding `tests` job to `windows.yml` will let us with:
- `windows.yml`: Runs `build` job and, if it succeeds, runs `tests` job.

So we still have the semantics we want and we still know which part has failed (has the build failed? has it succeded, but tests failed?) without duplication.

Coquinho:
> Just for clarification: having an additional `windows-test.yml` adds another "machine" to the CI that'll also build the EFL, so we'll have:
> 
> * `windows.yml`: Runs `build` job.
> * `windows-test.yml`: Runs `build` and then `tests` job.
> 
> Which means we'll be running the same `build` two times (and having to make sure that when one is changed, the other is also changed). That means that we're duplicating jobs here: if one `build` fails obviously the other `build` will fail too - it is _the same_ procedure.
> 
> By just adding `tests` job to `windows.yml` will let us with:
> 
> * `windows.yml`: Runs `build` job and, if it succeeds, runs `tests` job.
> 
> So we still have the semantics we want and we still know which part has failed (has the build failed? has it succeded, but tests failed?) without duplication.

Not exactly, `Jobs` at github actions don't share storage like that :/

JPTIZ:
> > Just for clarification: having an additional `windows-test.yml` adds another "machine" to the CI that'll also build the EFL, so we'll have:
> > 
> > * `windows.yml`: Runs `build` job.
> > * `windows-test.yml`: Runs `build` and then `tests` job.
> > 
> > Which means we'll be running the same `build` two times (and having to make sure that when one is changed, the other is also changed). That means that we're duplicating jobs here: if one `build` fails obviously the other `build` will fail too - it is _the same_ procedure.
> > By just adding `tests` job to `windows.yml` will let us with:
> > 
> > * `windows.yml`: Runs `build` job and, if it succeeds, runs `tests` job.
> > 
> > So we still have the semantics we want and we still know which part has failed (has the build failed? has it succeded, but tests failed?) without duplication.
> 
> Not exactly, `Jobs` at github actions don't share storage like that :/

Ok I actually discovered: they do work as expected _if you say one needs the other_, see [this test draft](https://github.com/expertisesolutions/efl/runs/842767015).

JPTIZ:
And e-mails stay like this:

JPTIZ:
And e-mails stay like this:

image

JPTIZ:
_Sheet_ now that I've read tests log: been refuted again

walac:
Adding tests to the same build job is a "no no" because then we wouldn't be able to determine if a PR broke the Windows build. I am in favor of making the test job dependent of the build.

Coquinho:
> Adding tests to the same build job is a "no no" because then we wouldn't be able to determine if a PR broke the Windows build. I am in favor of making the test job dependent of the build.

Didn't get it @walac, have a look at the checks for this PR. Even by making them dependent we can see if both have broken or only the tests.

walac:
> Didn't get it @walac, have a look at the checks for this PR. Even by making them dependent we can see if both have broken or only the tests.

I am talking about the suggestion of adding the test to windows.yml

Coquinho:
> > Didn't get it @walac, have a look at the checks for this PR. Even by making them dependent we can see if both have broken or only the tests.
> 
> I am talking about the suggestion of adding the test to windows.yml

@walac ,  we are doing it on `windows.yml` here.

PS: just notice a mistake I made on my last comment: by doing them in the same `.yml` we don't actually make then dependent as in github actions jobs don't share any storage.

Coquinho:
> Adding tests to the same build job is a "no no" because then we wouldn't be able to determine if a PR broke the Windows build. I am in favor of making the test job dependent of the build.

Oh just re-read it, I think you mean "independent" at the end right? If so, this PR is doing it.

JPTIZ:
> Adding tests to the same build job is a "no no" because then we wouldn't be able to determine if a PR broke the Windows build. I am in favor of making the test job dependent of the build.

But the suggestion wasn't to add to the same job, just in a different subsequent job depending on the build job. That's a common gitlab-CI use case, for example: https://docs.gitlab.com/ee/ci/pipelines/#group-jobs-in-a-pipeline

Coquinho:
Closing this in favor of #268

PR #266, opened 2020-07-06
[bin] Remove EAPI uses

Coquinho:
EAPI was being used at bin but as it isn't declared anymore its
generating some errors.


Coquinho:
This makes Linux work again too :D

cauebs:
> This makes Linux work again too :D

"All checks have passed" is something beautiful to read.

PR #268, opened 2020-07-06
Split CI jobs into "build" and "tests"

JPTIZ:
The behaviour should be close to #265, but with the build and the tests job splitted. This way we can see faster if it was the build or the tests that have failed.


.github/workflows/windows.yml
```diff
@@ -42,8 +40,6 @@ jobs:
       run: choco install nasm
     - name: Add Nasm to PATH
       run: echo "::add-path::C:\Program Files\Nasm"
-#    - name: test nasm
-#      run: nasm.exe -v
```
Coquinho, line 46: why?

JPTIZ: It was commented, but I could just uncomment if you prefer.

Coquinho: ohhh, right..duh...You can go on

PR #269, opened 2020-07-06
eo_test_general.c: Make eo_signals tests pass on Windows

walac:
EFL_CALLBACKS_ARRAY_DEFINE reorders the callbacks according to
efl_callbacks_cmp. efl_callbacks_cmp compares the address of the desc
field, which depends on the memory layout generated by the linker.

To make the test run deterministically, we define the array of callbacks
manually.

### DIFF
--- a/src/tests/eo/suite/eo_test_general.c
+++ b/src/tests/eo/suite/eo_test_general.c
@@ -199,11 +199,27 @@ _eo_signals_cb_added_deled(void *data EINA_UNUSED, const Efl_Event *event)
      fail_if(callback_array->func != _eo_signals_cb_added_deled);
 }
 
-EFL_CALLBACKS_ARRAY_DEFINE(_eo_signals_callbacks,
-{ EV_A_CHANGED, _eo_signals_a_changed_cb },
-{ EV_A_CHANGED, _eo_signals_a_changed_cb2 },
-{ EV_A_CHANGED, _eo_signals_a_changed_never },
-{ EFL_EVENT_DEL, _eo_signals_efl_del_cb });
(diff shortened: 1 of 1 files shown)


src/tests/eo/suite/eo_test_general.c
```diff
@@ -199,11 +199,27 @@ _eo_signals_cb_added_deled(void *data EINA_UNUSED, const Efl_Event *event)
...
+static Efl_Callback_Array_Item *
+_eo_signals_callbacks(void)
+{
+    static Efl_Callback_Array_Item items[] =
+      {
+          { EV_A_CHANGED, _eo_signals_a_changed_cb },
+          { EV_A_CHANGED, _eo_signals_a_changed_cb2 },
+          { EV_A_CHANGED, _eo_signals_a_changed_never },
+          { 0, _eo_signals_efl_del_cb },
+          { 0, 0 },
```
JPTIZ, line 213: What does the first `0` means here?

Also, maybe this change?
```suggestion
          { 0, _eo_signals_efl_del_cb },
          { 0, NULL },
```

Coquinho:
This change should only be applied to windows?

JPTIZ:
> This change should only be applied to windows?

I don't think so. This way it becomes independent of any linker's implementation, and also solves having this "duplicated".

src/tests/eo/suite/eo_test_general.c
walac: I am using `{ 0, 0 }` to make things exactly like the macro expansion does.

JPTIZ: I see, well, I think it is okay then.

PR #271, opened 2020-07-07
Fix openssl include path directory

walac:


joaoantoniocardoso:
This makes sense with new versions of meson, but what errors are you getting with openssl?

Coquinho:
This makes compilations to break for me.

joaoantoniocardoso:
> This makes compilations to break for me:
> 
> ```
> 3 warnings generated.                                   
> [471/568] Linking target src/lib/emile/emile-1.dll       
> FAILED: src/lib/emile/emile-1.dll src/lib/emile/emile-1.p
> db                                                      
> lld-link @src/lib/emile/emile-1.dll.rsp                 
> lld-link: error: could not open 'libssl_static.lib': no s
> uch file or directory                                   
> lld-link: error: could not open 'libcrypto_static.lib': n
> o such file or directory                                
> [492/568] Compiling C objec...ce@exe/interface_main.c.obj
> ninja: build stopped: subcommand failed.                
> ninja: Entering directory `C:\Users\Lucas\source\repos\ef
> l2\build'                                               
> [2/77] Linking target src/lib/emile/emile-1.dll          
> FAILED: src/lib/emile/emile-1.dll src/lib/emile/emile-1.p
> db                                                      
> lld-link @src/lib/emile/emile-1.dll.rsp                 
> lld-link: error: could not open 'libssl_static.lib': no s
> uch file or directory                                   
> lld-link: error: could not open 'libcrypto_static.lib': n
> o such file or directory                                
> [23/77] Compiling C object ...exe/constructors_main.c.obj
> ninja: build stopped: subcommand failed.                
> Could not rebuild C:\Users\Lucas\source\repos\efl2\build 
> ```
> 
> Is it expected?

Oh, I see, could you try removing that extra `/ 'lib'` added?

walac:
> This makes sense with new versions of meson, but what errors are you getting with openssl?

The include files are not found.

joaoantoniocardoso:
> > This makes sense with new versions of meson, but what errors are you getting with openssl?
> 
> The include files are not found.

Could you check if is this [same issue](https://github.com/expertisesolutions/efl/pull/250#issuecomment-653685434)?

walac:
> Could you check if is this same issue?

Yeah, the problem is that I use a prebuilt version of OpenSSL, where the lib files are inside a `/lib` subdirectory.

meson.build
```diff
@@ -340,7 +340,7 @@ elif (get_option('crypto') == 'gnutls')
 elif (get_option('crypto') == 'openssl')
   if (get_option('openssl_dir') != '')
     crypto = declare_dependency(
-      link_args: ['-L' + get_option('openssl_dir'), '-llibssl_static', '-llibcrypto_static'],
+      link_args: ['-L' + get_option('openssl_dir'), '-L' + get_option('openssl_dir') + '/lib', '-llibssl_static', '-llibcrypto_static'],
```
JPTIZ, line 343: I think it should be `get_option('openssl_dir') / 'lib'`, but...

walac: It feels like the CI has a version of meson that doesn't support the syntax.

JPTIZ: Windows CI version is always the updated from PyPI (0.54.3), so it should support. Didn't check Travis, tho

PR #273, opened 2020-07-07
Building eet

joaoantoniocardoso:
As commented in #138, building eet is straight-forward, but its tests is failing:
```
24/24 eet-suite                 FAIL           12.93s (exit status 1)
```


JPTIZ:
Done, now we finally **_enabled it_**

PR #275, opened 2020-07-07
[efl mono] Define and use EFL_MONO_API

Coquinho:
1. Create `lib_api.h`;
   `lib_api.h`;
8. Add `LIB_api.h` to public headers;
9. If files are generated here, add `-e` flag to `eolian_gen` with
   `LIB_API` as argument (It didn't happen until now);

definition of `LIB_API`.

This one also worked on efl_mono tests, so `EAPI`/`EWAPI`/`EOAPI` were removed,
where those were defined, now I define `EFL_MONO_TEST_API` and its `WEAK`
contra part. Also, `meson.build` should change accordingly to pass `-e
EFL_MONO_TEST_API`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,8 +46,17 @@ env:
 
 jobs:
   include:
+    - os: osx
+    - os: linux
+      env: DISTRO=Fedora32-mingw CI_BUILD_TYPE=mingw
+    - os: linux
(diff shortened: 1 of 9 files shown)


Coquinho:
Yay, with every CI enable it falls on #144 again. 
I will remove other Travis-CI things and let this ready.

PR #276, opened 2020-07-07
Building Ector

joaoantoniocardoso:
This PR is redoing #139, but merging to native-windows.

Ector is building and passing tests:
```
25/25 ector-suite               OK             0.11s
```

Note that this PR depends on (#273).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -506,7 +506,6 @@ if sys_windows
     'ecore_win32',
     'ecore_wl2',
     'ecore_x',
-    'ector',
     'edje',
     'eeze',
     'efl_canvas_wl',


meson.build
```diff
@@ -506,9 +506,7 @@ if sys_windows
     'ecore_win32',
     'ecore_wl2',
     'ecore_x',
-    'ector',
     'edje',
-    'eet',
```
Coquinho, line 511: [ector doesn't need eet](https://github.com/expertisesolutions/efl/blob/devs/joaoantoniocardoso/ector/redone/meson.build#L456) so you don't need to make um PR dependent on the other.

joaoantoniocardoso:
> [ector doesn't need eet](https://github.com/expertisesolutions/efl/blob/devs/joaoantoniocardoso/ector/redone/meson.build#L456) so you don't need to make a PR dependent on the other.

It is not specified in `src/meson.build` but it is specified [here](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/src/lib/ector/meson.build#L1).

Coquinho:
> > [ector doesn't need eet](https://github.com/expertisesolutions/efl/blob/devs/joaoantoniocardoso/ector/redone/meson.build#L456) so you don't need to make a PR dependent on the other.
> 
> It is not specified in `src/meson.build` but it is specified [here](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/src/lib/ector/meson.build#L1).

Ohh, my bad

Coquinho:
@joaoantoniocardoso could you rebase with native-windows so we have more Travis here, please?

PR #277, opened 2020-07-07
[WIP] [Bug] An example was using EAPI instead of EAPI_MAIN as all the others

Coquinho:
This is the only `EAPI` on `src/example` beside it, there is `EAPI_MAIN` (which is defined by core) so I'm following the pattern that this `EAPI` should be `EAPI_MAIN`.


Coquinho:
(rebase with native-windows)

Coquinho:
IT WORKS, vou reverter aqui as outras CI.

PR #283, opened 2020-07-09
[efl] Change efl build definition

Coquinho:
As `EFL_BUILD` is already a thing, this changes `EFL_BUILD` to
`EFL_LIB_BUILD` removing any chance of wrongly setting it.

Based on [946db](https://github.com/expertisesolutions/efl/commit/946db530a7dc799b841f24311b68a23ebc439497)

### DIFF
--- a/src/lib/efl/efl_api.h
+++ b/src/lib/efl/efl_api.h
@@ -7,7 +7,7 @@
 
 #ifdef _WIN32
 # ifndef EFL_STATIC
-#  ifdef EFL_BUILD
+#  ifdef EFL_LIB_BUILD
 #   define EFL_API __declspec(dllexport)
 #  else
 #   define EFL_API __declspec(dllimport)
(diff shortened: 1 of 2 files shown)


Coquinho:
> Isn't there tests as well?

Tests don't need to be exported/imported.

> Y'know, the `-e` stuff.

no no, this is only to know if we need to import or export at `efl_api.h` this is not the `LIB_API` definition purse.

JPTIZ:
> > Isn't there tests as well?
> 
> Tests don't need to be exported/imported.
> 
> > Y'know, the `-e` stuff.
> 
> no no, this is only to know if we need to import or export at `efl_api.h` this is not the `LIB_API` definition purse.

PFFF sorry I messed up things.

It is about generated eo's I was thinking about. You know, the `-e` stuff². But you're saying it doesn't apply to `lib/efl` in this case, right? Then...

JPTIZ:
Right right right I heavily mixed everything up. This is about `_BUILD` and not about `_API`, herp.

PR #286, opened 2020-07-09
evas: Add guards to some includes

Coquinho:
This is based on [805dccea](https://github.com/expertisesolutions/efl/commit/805dccea1434e7bc6bbf1d112b29ba58063475e0) but instead of guarding with `_WIN32` I'm guarding it with `_MSC_VER` as this should already work with `msys`

### DIFF
--- a/src/lib/evas/common/evas_image_load.c
+++ b/src/lib/evas/common/evas_image_load.c
@@ -4,7 +4,14 @@
 
 #include <sys/types.h>
 #include <sys/stat.h>
-#include <unistd.h>
+
+#ifndef _MSC_VER
+# include <unistd.h>
+#endif
+
+#ifdef _WIN32_
+# include <evil_private.h>
+#endif
 
 #include "evas_common_private.h"
 #include "evas_private.h"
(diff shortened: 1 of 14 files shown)


src/lib/evas/include/evas_common_private.h
```diff
@@ -22,16 +22,22 @@
...
 #include <stdint.h>
-#include <unistd.h>
+
+#ifndef _MSC_VER
+# include <unistd.h>
+# include <sys/types.h>
+# include <sys/stat.h>
+#endif
+
+#ifndef _WIN32
```
Coquinho, line 32: whops

src/lib/evas/common/evas_image_load.c
```diff
@@ -23,43 +30,43 @@ struct ext_loader_s
 static const struct ext_loader_s loaders[] =
 { /* map extensions to loaders to use for good first-guess tries */
    MATCHING(".png", "png"),
-   
+
    MATCHING(".jpg", "jpeg"),
    MATCHING(".jpeg", "jpeg"),
    MATCHING(".jfif", "jpeg"),
-   
+
```
JPTIZ, line 37: ` `

src/tests/evas/evas_test_image.c
```diff
@@ -5,7 +5,14 @@
 #include <stdio.h>
 #include <sys/types.h>
 #include <sys/stat.h>
-#include <unistd.h>
+
+#ifndef _MSC_VER
+# include <unistd.h>
+#endif
+
+#ifndef _WIN32
```
Coquinho, line 13: whops

Coquinho:
> Ctrl+F for "ifndef _W" returns no result, so I think it might be ok.

Travis id ok, so we may be ok too.

PR #291, opened 2020-07-10
[wrap] Fixing jpeg

joaoantoniocardoso:


meson.build
```diff
@@ -60,6 +52,31 @@ sys_bsd = bsd.contains(host_machine.system())
 sys_windows = windows.contains(host_machine.system())
 sys_osx = osx.contains(host_machine.system())
 
+if sys_windows
+  nuget_download = get_option('nuget-download')
+  if not nuget_download
+    message('needs -Dnuget-download=true to download deps')
+  endif
+  if cc.has_header_symbol('stdio.h', '_MSC_VER') and nuget_download
```
Coquinho, line 60: This could be:
```suggestion
  elif cc.has_header_symbol('stdio.h', '_MSC_VER') 
```

meson.build
```diff
@@ -60,6 +52,31 @@ sys_bsd = bsd.contains(host_machine.system())
...
+    endif
+
+    message(output)
+
+    subdir('win32-deps')
+  else
+    message('no nuget download because download failed or something')
+  endif
+else
+  message('not on windows')
```
Coquinho, line 77: Do we need this message?

meson.build
```diff
@@ -60,6 +52,31 @@ sys_bsd = bsd.contains(host_machine.system())
...
+    if r.returncode() != 0
+      errortxt = r.stderr().strip()
+      error(errortxt)
+    endif
+
+    message(output)
+
+    subdir('win32-deps')
+  else
+    message('no nuget download because download failed or something')
```
Coquinho, line 74: Could this be a warning?

meson.build
joaoantoniocardoso: I think not, nuget should only be downloaded (line below) if `nuget_download` is set to true.

meson.build
joaoantoniocardoso: not at all.

meson.build
Coquinho: Yeah, how it is now it doesn't make sense.

meson.build
```diff
@@ -60,6 +52,27 @@ sys_bsd = bsd.contains(host_machine.system())
...
+    endif
+
+    message(output)
+
+    subdir('win32-deps')
+  else
+    message('nuget-download disabled, if you want to use it to get dependencies you need to pass -Dnuget-download=true or set it in native-file-windows.txt')
+  endif
+endif
+
```
cauebs, line 75: Is this still necessary if we're using a wrap?

joaoantoniocardoso: At the point of this PR it was necessary to maintain Nuget working supposedly for the future. Turns out that we decided today to remove Nuget and then #334 does exactly that.

PR #292, opened 2020-07-10
[wrap] lua

joaoantoniocardoso:


meson.build
```diff
@@ -395,13 +395,13 @@ if sys_windows or get_option('lua-interpreter') == 'lua'
...
-      break
-    endif
-  endforeach
+  #foreach l : luaold_interpreters
+  #  lua = dependency(l[0], version: l[1], required:false)
+  #  lua_pc_name = l[0]
+  #  if lua.found() == true
+  #    break
+  #  endif
+  #endforeach
```
Coquinho, line 404: This should not run on linux neither?

joaoantoniocardoso: Possibly yes as it is still in master, I'll guard it to windows only.

meson.build
```diff
@@ -389,22 +389,27 @@ luaold_interpreters = [
...
-    warning('There is no "' + get_option('lua-interpreter')
-            + '" binding option for Windows, backing up to "lua"')
-  endif
-  config_h.set('ENABLE_LUA_OLD', '1')
-  foreach l : luaold_interpreters
-    lua = dependency(l[0], version: l[1], required:false)
-    lua_pc_name = l[0]
-    if lua.found() == true
-      break
+if not lua.found():  # see header_checks/meson.build
```
JPTIZ, line 392: This will cause an error on Linux, since `lua` was never defined.

JPTIZ: Also, we're changing the behaviour here. What if our backend is actually luajit?

meson.build
```diff
@@ -389,22 +389,27 @@ luaold_interpreters = [
...
+    foreach l : luaold_interpreters
+      lua = dependency(l[0], version: l[1], required:false)
+      lua_pc_name = l[0]
+      if lua.found() == true
+        break
+      endif
+    endforeach
+    if not lua.found()
+      error('Lua not found')
+    endif
```
JPTIZ, line 408: My recommendation is to change here:

```suggestion
    if not lua.found()
      lua = dependency('lua', fallback : ['lua', 'lua_dep'])
      if not lua.found()
        error('Lua not found')
      endif
    endif
```

joaoantoniocardoso:
Ok, I set `lua-interpreter` to `lua` on native file but it keeps with `luajit`, any clue @JPTIZ ?

Coquinho:
> Ok, I set `lua-interpreter` to `lua` on native file but it keeps with `luajit`, any clue @JPTIZ ?

@joaoantoniocardoso I think you should change `meson_options.txt`, there lua-interpreter is being set to luajit.

JPTIZ:
> 
> 
> Ok, I set `lua-interpreter` to `lua` on native file but it keeps with `luajit`, any clue @JPTIZ ?

Native-files are still not setting options, actually (see https://github.com/mesonbuild/meson/issues/6783). You'll have to set it manually through command-line:

```console
$ meson build --reconfigure -Dlua-interpreter=lua
```


> 
> 
> > Ok, I set `lua-interpreter` to `lua` on native file but it keeps with `luajit`, any clue @JPTIZ ?
> 
> @joaoantoniocardoso I think you should change `meson_options.txt`, there lua-interpreter is being set to luajit.

That's the default, we shouldn't have to change it there.

joaoantoniocardoso:
> > Ok, I set `lua-interpreter` to `lua` on native file but it keeps with `luajit`, any clue @JPTIZ ?
> 
> Native-files are still not setting options, actually (see [mesonbuild/meson#6783](https://github.com/mesonbuild/meson/issues/6783)). You'll have to set it manually through command-line:
> 
> ```
> $ meson build --reconfigure -Dlua-interpreter=lua
> ```
Oh boy, we are out of line-length for additional parameters. If this is the case, why does nuget-download on native-file works?

Coquinho:
> > Ok, I set `lua-interpreter` to `lua` on native file but it keeps with `luajit`, any clue @JPTIZ ?
> 
> Native-files are still not setting options, actually (see [mesonbuild/meson#6783](https://github.com/mesonbuild/meson/issues/6783)). You'll have to set it manually through command-line:
> 
> ```
> $ meson build --reconfigure -Dlua-interpreter=lua
> ```

So we could add this to configure.bat, right?
 
> > > Ok, I set `lua-interpreter` to `lua` on native file but it keeps with `luajit`, any clue @JPTIZ ?
> > 
> > 
> > @joaoantoniocardoso I think you should change `meson_options.txt`, there lua-interpreter is being set to luajit.
> 
> That's the default, we shouldn't have to change it there.

It was changed by us actually (I git blamed it), that's why I thought of it.

JPTIZ:
> Oh boy, we are out of line-length for additional parameters. If this is the case, why does nuget-download on native-file works?

So, I realized it's actually the length of __the variable__ that is the problem. I'll quickly submit a PR for additional flags. But even so, a `--reconfigure` will not come from `configure` :) Run `configure.bat` and _then_ you run that meson command. It'll just change what is the the definition of `lua-interpreter`.

JPTIZ:
> 
> 
> > > Ok, I set `lua-interpreter` to `lua` on native file but it keeps with `luajit`, any clue @JPTIZ ?
> > 
> > 
> > Native-files are still not setting options, actually (see [mesonbuild/meson#6783](https://github.com/mesonbuild/meson/issues/6783)). You'll have to set it manually through command-line:
> > ```
> > $ meson build --reconfigure -Dlua-interpreter=lua
> > ```
> 
> So we could add this to configure.bat, right?

I don't think it is necessary, since it is a matter of installing luajit. But maybe for a matter of defaults? Hm...

> > > > Ok, I set `lua-interpreter` to `lua` on native file but it keeps with `luajit`, any clue @JPTIZ ?
> > > 
> > > 
> > > @joaoantoniocardoso I think you should change `meson_options.txt`, there lua-interpreter is being set to luajit.
> > 
> > 
> > That's the default, we shouldn't have to change it there.
> 
> It was changed by us actually (I git blamed it), that's why I thought of it.

:P

joaoantoniocardoso:
> > That's the default, we shouldn't have to change it there.
> 
> It was changed by us actually (I git blamed it), that's why I thought of it.

Oh, our default could change, though.


> > Oh boy, we are out of line-length for additional parameters. If this is the case, why does nuget-download on native-file works?
> 
> So, I realized it's actually the length of **the variable** that is the problem. I'll quickly submit a PR for additional flags. But even so, a `--reconfigure` will not come from `configure` :) Run `configure.bat` and _then_ you run that meson command. It'll just change what is the the definition of `lua-interpreter`.

Nice catch, then one could loop for a list of arguments doing `--reconfigure`?

JPTIZ:
> 
> 
> > > That's the default, we shouldn't have to change it there.
> > 
> > 
> > It was changed by us actually (I git blamed it), that's why I thought of it.
> 
> Oh, our default could change, though.
> 
> > > Oh boy, we are out of line-length for additional parameters. If this is the case, why does nuget-download on native-file works?
> > 
> > 
> > So, I realized it's actually the length of **the variable** that is the problem. I'll quickly submit a PR for additional flags. But even so, a `--reconfigure` will not come from `configure` :) Run `configure.bat` and _then_ you run that meson command. It'll just change what is the the definition of `lua-interpreter`.
> 
> Nice catch, then one could loop for a list of arguments doing `--reconfigure`?

I thought about but it actually would be the worse of two words haha Remember: meson regenerates build dir _every time_ you make a `--reconfigure`. And sometimes changes may end not acting properly. Just splitting into two variables (if it is the case of new args) would be enough.

joaoantoniocardoso:
> > > > That's the default, we shouldn't have to change it there.
> > > 
> > > 
> > > It was changed by us actually (I git blamed it), that's why I thought of it.
> > 
> > 
> > Oh, our default could change, though.
> > > > Oh boy, we are out of line-length for additional parameters. If this is the case, why does nuget-download on native-file works?
> > > 
> > > 
> > > So, I realized it's actually the length of **the variable** that is the problem. I'll quickly submit a PR for additional flags. But even so, a `--reconfigure` will not come from `configure` :) Run `configure.bat` and _then_ you run that meson command. It'll just change what is the the definition of `lua-interpreter`.
> > 
> > 
> > Nice catch, then one could loop for a list of arguments doing `--reconfigure`?
> 
> I thought about but it actually would be the worse of two words haha Remember: meson regenerates build dir _every time_ you make a `--reconfigure`. And sometimes changes may end not acting properly. Just splitting into two variables (if it is the case of new args) would be enough.

hahah yeah you're right, I can see the huge mess it would be. But for now isn't better to just revert the default to `lua`?

joaoantoniocardoso:
Linux CI didn't like that `lua` `lua-interpreter`

JPTIZ:
> 
> 
> Linux CI didn't like that `lua` `lua-interpreter`. Master is using `luajit` too.

lol you're right, master uses `luajit`. Well, then let it as is.

header_checks/meson.build
```diff
@@ -221,6 +221,11 @@ endif
 # libcheck
 check = dependency('check', fallback : ['check', 'check_dep'])
 
+if sys_windows
+  lua_project = subproject('lua')
+  lua = lua_project.get_variable('lua_dep')
+endif
+
```
JPTIZ, line 228: This is not necessary anymore (due to that change with the `fallback` right there below.

cauebs:
On Linux I still get `src/bindings/meson.build:27:4: ERROR: Non-existent build file 'src/bindings/luajit/meson.build'`

Maybe this is a good time to fix that?

I always do the following on `src/bindings/meson.build`

```diff
-
 bindings = get_option('bindings')
-if not sys_windows
-  bindings_order = ['luajit', 'cxx', 'mono']
-else
-  bindings_order = ['lua', 'cxx', 'mono']
-endif
+bindings_order = ['lua', 'cxx', 'mono']
```

meson.build
```diff
@@ -389,7 +389,7 @@ luaold_interpreters = [
 
 lua_pc_name = ''
 
-if sys_windows or get_option('lua-interpreter') == 'lua'
+if get_option('lua-interpreter') == 'lua'
   if sys_windows and get_option('lua-interpreter') != 'lua'
     warning('There is no "' + get_option('lua-interpreter')
             + '" binding option for Windows, backing up to "lua"')
```
Coquinho, line 395: Is this still needed?

joaoantoniocardoso: Yes, it is showing a warning in the case of someone not changing the default `lua-interpter` (that is `luajit`) to `lua`.

Coquinho: But if that happens there is no way he enters this `if`

joaoantoniocardoso: oooh thanks, I'll fix that

meson.build
```diff
@@ -394,7 +390,17 @@ if sys_windows or get_option('lua-interpreter') == 'lua'
...
+    lua = dependency('lua', fallback : ['lua', 'lua_dep'])
+    if not lua.found()
+      error('Lua not found')
+    endif
+  endif
 else
+  if sys_windows 
+    warning('There is no "' + get_option('lua-interpreter')
+            + '" binding option for Windows, backing up to "lua"')
+  endif
```
Coquinho, line 403: Where is this backup happening? If we want to do this backup thing, why not let how it was before this PR (lines 384~388)?

meson.build
```diff
@@ -394,17 +402,14 @@ if sys_windows or get_option('lua-interpreter') == 'lua'
...
   endforeach
+  if not lua.found()
+    lua = dependency('lua', fallback : ['lua', 'lua_dep'])
+    if not lua.found()
+      error('Lua not found')
+    endif
+  endif
 else
   lua = dependency(get_option('lua-interpreter'))
-  lua_pc_name = 'luajit'
```
Coquinho, line 399: Are there any other options for lua other than `lua` and `luajit`?

meson.build
```diff
@@ -381,11 +381,19 @@ luaold_interpreters = [
 
 lua_pc_name = ''
 
-if sys_windows or get_option('lua-interpreter') == 'lua'
-  if sys_windows and get_option('lua-interpreter') != 'lua'
-    warning('There is no "' + get_option('lua-interpreter')
-            + '" binding option for Windows, backing up to "lua"')
+if get_option('lua-interpreter') == 'luajit'
+  if sys_osx == true
```
Coquinho, line 385: If we are touching this, why not letting it more python-like:
```suggestion
  if sys_osx
```

meson.build
```diff
@@ -381,11 +381,19 @@ luaold_interpreters = [
...
+    # luajit on macro is broken, this means we need to generate our own dependency with our arguments, a library later still needs to link to luajit for the pagesize argument thingy
+    lua = declare_dependency(
+      include_directories: include_directories(lua.get_pkgconfig_variable('includedir')),
+      link_args: ['-L'+lua.get_pkgconfig_variable('libdir'),
+                  '-l'+lua.get_pkgconfig_variable('libname')]
+    )
+  else
+    lua = dependency('luajit', fallback : ['luajit', 'luajit_dep'])
```
Coquinho, line 393: This defines a `luajit` wrap, right? If so there should be a `luajit.warp` doesn't it?

meson.build
joaoantoniocardoso: Currently `lua-interpreter` can be `luajit` (default) or `lua`, and then `lua_pc_name` can be `luajit`, `lua`, `lua51`, `lua-5.1`, `lua5.1`, `lua52`,  `lua-5.2` or `lua5.2'`.
I haven't tried to set `lua-interpreter` to something different from what is specified at `meson_options.txt` but I expect meson to prevent that, so if that is the case we could remove the else part completly.

Coquinho: I think we can remove this then.

meson.build
joaoantoniocardoso: Yeah, if luajit is not found in the system, it will fall back to the subproject called luajit (`subprojects/luajit.wrap`). If not available I guess it will show a warning. The lua_pc_name will propagate the 'luajit' name for the whole project.

Do you think it is not the desired behavior or maybe I missed something..?

Coquinho: There is no `luajit.wrap` in this PR. At least I can't find it.

joaoantoniocardoso: Yeah, we can remove the fallback. It will be added with 'luajit' wrapper if that is the case.
```suggestion
    lua = dependency('luajit')
```

Coquinho: It would make more sense to me.

felipealmeida:
We should probably just keep what is on Linux to Windows too and add fallbacks

felipealmeida:
With vcpkg maybe not even that

meson.build
```diff
@@ -380,30 +380,52 @@ luaold_interpreters = [
...
+  endif
 else
   lua = dependency(get_option('lua-interpreter'))
   lua_pc_name = 'luajit'
+  if not lua.found()
+    lua = dependency('luajit')
+    if not lua.found()
+      error('Lua not found')
+    endif
+  endif
```
Coquinho, line 419: As for now we don't have a `luajit` fallback, so I think we don't need to double check it:
```suggestion
else
  lua = dependency(get_option('lua-interpreter'))
  lua_pc_name = 'luajit'
  if not lua.found()
    error('Lua not found')
  endif
```

meson.build
```diff
@@ -380,30 +380,49 @@ luaold_interpreters = [
...
     endif
   endforeach
+  if not lua.found()
+    lua = dependency('lua', fallback : ['lua', 'lua_dep'])
+    if not lua.found()
+      error('Lua not found')
+    endif
+  endif
+  if have_elua
+    luaver_min = cc.compute_int('LUA_VERSION_NUM - 500',
```
cauebs, line 401: If I enable elua, I get an error here. 
```
meson.build:401:4: ERROR: Dependencies must be external dependencies
```

joaoantoniocardoso: Yep, `Elua` is disabled in this PR. I tried to enabled it using `lua-interpreter=luajit` and worked, but probably we should not build `elua` as it is disabled by default in master.

cauebs: But then how do you check if your changes are correct?

joaoantoniocardoso: Oh, right, try to enable `Evas`, as it is done in #332 .

cauebs: Well, it builds on Linux!

PR #293, opened 2020-07-10
[Evil] Implementing dirname

joaoantoniocardoso:
This PR aims to implement the missing `dirname` in some way that it should work when using multiple threads.


src/lib/evil/evil_dirname.c
```diff
@@ -0,0 +1,16 @@
+#ifdef HAVE_CONFIG_H
+# include "config.h"
+#endif /* HAVE_CONFIG_H */
+
+#include <evil_api.h>
```
walac, line 5: nit: this was already included in `evil_dirname.h`

src/lib/evil/evil_dirname.c
```diff
@@ -0,0 +1,16 @@
...
+
+#include <stdlib.h>
+
+#include <evil_windows.h>
+
+EVIL_API char* dirname (char* path)
+{
+  static char dir[MAX_PATH];
```
walac, line 13: Use a threa local storage, otherwise this can't be called from multiple threads.

joaoantoniocardoso: Oh then we should fix evil's `basename()` too.

src/lib/evil/evil_libgen.h
```diff
@@ -0,0 +1,13 @@
+#ifndef __EVIL_LIBGEN_H__
+#define __EVIL_LIBGEN_H__
+
+#include <evil_private.h>
+
+#ifdef _MSC_VER
+# define dirname evil_dirname
+#endif
```
Coquinho, line 8: I bring the same question that I brought at `evil_basename`, should we use [_MSC_VER or HAVE_FOO](https://github.com/expertisesolutions/efl/pull/323#discussion_r458078126) and why?

PR #294, opened 2020-07-10
[Deps] [Redone] Ecore without tests

joaoantoniocardoso:
After #159 Ecore will be building fine **without tests**.

However before this PR is ready for review we need to:
  - [x] gettimeofday into native-windows (fixed by #159)
  - [x] properly fix winsock2 (fixed by #197)

To be clear why **without tests**, to enable test to be built we'll need to build:
  - ecore_file  
     - ecore_con  
  - ecore_input  
  - ecore_evas  
    - ecore_input_evas  
       - evas  
       - ecore_win32
  - ecore_imf  
  - _and possibly others..._

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -439,7 +439,7 @@ subprojects = [
 ['eet'              ,[]                    , false,  true,  true, false,  true,  true,  true, ['eina', 'emile', 'efl'], []],
 ['ecore'            ,[]                    , false,  true, false, false, false, false,  true, ['eina', 'eo', 'efl'], ['buildsystem']],
 ['eldbus'           ,[]                    , false,  true,  true, false,  true,  true,  true, ['eina', 'eo', 'efl'], []],
-['ecore'            ,[]                    ,  true, false, false, false,  true,  true,  true, ['eina', 'eo', 'efl'], []], #ecores modules depend on eldbus
+['ecore'            ,[]                    ,  true, false, false, false, false,  true,  true, ['eina', 'eo', 'efl'], []], #ecores modules depend on eldbus
 ['ecore_audio'      ,['audio']             , false,  true, false, false, false, false,  true, ['eina', 'eo'], []],
 ['ecore_avahi'      ,['avahi']             , false,  true, false, false, false,  true, false, ['eina', 'ecore'], []],
 ['ecore_con'        ,[]                    , false,  true,  true, false,  true, false,  true, ['eina', 'eo', 'efl', 'ecore'], ['http-parser']],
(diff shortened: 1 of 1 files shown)


Coquinho:
> To be clear why **without tests**, to enable test to be built we'll need to build:
> 
> * ecore_file
>   
>   * ecore_con
> * ecore_input
> * ecore_evas
>   
>   * ecore_input_evas
>     
>     * evas
>     * ecore_win32
> * ecore_imf
> * _and possibly others..._

evas-fixes has ecore tests running, so it won't be a long wait.

src/lib/ecore/ecore_time.c
```diff
@@ -55,14 +55,14 @@ ecore_time_get(void)
...
    struct timeval timev;
 
    gettimeofday(&timev, NULL);
    return (double)timev.tv_sec + (((double)timev.tv_usec) / 1000000);
-#else
-# error "Your platform isn't supported yet"
-#endif
+//#else
+//# error "Your platform isn't supported yet"
+//#endif
```
Coquinho, line 65: Instead of commenting all this, you could make this be a warning instead of an error.

src/lib/evil/evil_strings.h
```diff
@@ -2,14 +2,9 @@
...
+#include <evil_windows.h>
 
-#ifdef _WIN32
-# ifndef WIN32_LEAN_AND_MEAN
-#  define WIN32_LEAN_AND_MEAN
-# endif
-# include <windows.h>
-# undef WIN32_LEAN_AND_MEAN
-#else
+#ifndef _WIN32
```
Coquinho, line 7: This probably should be `_MSC_VER` as evil only runs on windows.

joaoantoniocardoso:
As soon as #159 merges the ecore will be building.

src/lib/ecore/ecore_time.c
```diff
@@ -61,7 +61,7 @@ ecore_time_unix_get(void)
    gettimeofday(&timev, NULL);
    return (double)timev.tv_sec + (((double)timev.tv_usec) / 1000000);
 #else
-# error "Your platform isn't supported yet"
+# warning "Your platform isn't supported yet"
```
Coquinho, line 64: why is this change needed?

joaoantoniocardoso: I maybe got it wrong but wasn't your [suggestion](https://github.com/expertisesolutions/efl/pull/294#discussion_r453602013)?

Coquinho: for while `gettimeofday` wasn't being implemented, yes. Now that we have it and we define `HAVE_GETTIMEOFDAY` there is no reason to maintain this change.

PR #301, opened 2020-07-13
win32: Remove multiple frees for stringshare

Coquinho:
The function available_types do not get ownership of the acceptable_types, and so shouldn't free the stringshares it receives. Doing so causes double free.

Original from https://github.com/expertisesolutions/efl/commit/60c3f1e8a5a8ad2c052c00a4be3bda450d781514


Coquinho:
> The function available_types do not get ownership of the acceptable_types, and so shouldn't free the stringshares it receives. Doing so causes double free. <- This can be included in the commit message?

I've added it to the PR message, when the time to `squash and merge` arrives I'll put on the commit.

PR #302, opened 2020-07-13
win32: Fix unistd.h include and incdir

Coquinho:
Based on https://github.com/expertisesolutions/efl/commit/f199fbc671f0827999eccc283734ea9d7b98defe

### DIFF
--- a/src/generic/evas/meson.build
+++ b/src/generic/evas/meson.build
@@ -12,7 +12,8 @@ subdir('common')
 inc_dir = config_dir
 common_deps = generic_deps
 if sys_windows
-  inc_dir = [inc_dir, zlib_include_dir]
+  inc_dir = [inc_dir]
+  common_deps = [evil, zlib]
 else
   common_deps = [generic_deps, rt]
 endif


src/generic/evas/meson.build
```diff
@@ -12,7 +12,8 @@ subdir('common')
 inc_dir = config_dir
 common_deps = generic_deps
 if sys_windows
-  inc_dir = [inc_dir, zlib_include_dir]
+  inc_dir = [inc_dir]
+  common_deps = [evil, zlib]
```
JPTIZ, line 16: doesn't evas depend on `eina`? If so, `evil` should be unnecessary here.

Coquinho: @JPTIZ how can I check if generic evas already depends on eina?

JPTIZ: @Coquinho The non-quick-but-easy is to track dependencies until one of them depends on Eina. But almost every EFL subproject should already depend on eina, hm...

The other way is to use `meson introspect --targets -i <builddir> | less` and check if `src/lib/eina` (or you could directly search for `src/lib/evil`) is on any `-I` flag.

Coquinho: Tried to [remove evil](https://github.com/expertisesolutions/efl/commit/ac4382ad9132d2fcaf11f389d3370e254165f78e), got this [error  ](https://github.com/expertisesolutions/efl/runs/870622102#step:23:412) about not being able to include `evil_private`. It seems that neither eina nor evil is being included here.

JPTIZ: Ok, now it makes sense: it didn't crash because we didn't _have_ `evil_private.h`, and `generic/evas/common` for some reason didn't depend on anywone (why? no idea). We introduce `evil_private.h` here: https://github.com/expertisesolutions/efl/compare/devs/coquinho/clean/evas-fixes#diff-fea4a2f39268ddbfeddb4ed4edaf1c38R23.

Ok, in this case it is 100% fair to depend on `evil`.

PR #305, opened 2020-07-13
Install ninja through pip on Windows CI

cauebs:
This shaves off at least a minute from the CI builds :rocket: 

Meson and Ninja are Python applications. Installing through choco probably required downloading a much beefier package, with an embedded CPython build. Installing with pip is the logical way to do it, especially considering we already install Meson this way.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -44,7 +44,7 @@ jobs:
       run: nasm -v
     - name: install ninja
       # unexplicably, installation returns error code 1 if a cache location is used
-      run: choco install ninja
+      run: py -3 -m pip install ninja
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -42,18 +42,17 @@ jobs:
       run: echo "::add-path::C:\Program Files\Nasm"
     - name: Test nasm
       run: nasm -v
+    - name: add python scripts to path
```
cauebs, line 45: Just moved this up before the ninja installation and renamed a variable.

cauebs: It's replicated for the test job, below.

PR #306, opened 2020-07-14
build: Enable sending extra args to configure.bat

JPTIZ:
Enables running \`configure \<args\>\` (e.g. \`configure -Dlua-interpreter=luajit\`)

### DIFF
--- a/configure.bat
+++ b/configure.bat
@@ -8,6 +8,8 @@ if "%~1"=="--verbose" (
     echo Verbose OFF.
 )
 
+set MESONFLAGS_EXTRA=%*
+
 call :main || (echo Build configure failed.)
 exit /B %errorlevel%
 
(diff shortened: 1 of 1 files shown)


configure.bat
```diff
@@ -108,18 +110,21 @@ exit /B 0
...
+    @echo Extra flags: %MESONFLAGS_EXTRA%
+    @echo ------------------------------
+    @set MESONFLAGS=%MESONFLAGS:            = %
 exit /B 0
 
 :generate_build
     @echo ------------------------------
     @echo Generating build...
-    set vcvars64="C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat"
+    set vcvars64="C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat"
```
Coquinho, line 121: This change is needed?

JPTIZ: PFF WOW Accidental actually :D Maybe we should check for VS version when doing this, hm...

Coquinho: Yeah, seems like a good idea

PR #312, opened 2020-07-15
evil: Define sleep

Coquinho:
Define sleep function and HAVE_SLEEP macro for windows builds if sleep is not already
defined.


felipealmeida:
Commit com modificações de espaço?

Coquinho:
> Commit com modificações de espaço?

Some of them are removing of trailing spaces, but as `usleep` was defined by us I don't see why we should add more trailing spaces to EFL.
The others are `CRFL` vs `LF` so it makes sense.

PR #314, opened 2020-07-15
win32: Add eet dependency  to vg_common

Coquinho:


JPTIZ:
Do you know which problem was this solving exactly?

Coquinho:
> Do you know which problem was this solving exactly?

It couldn't find `Eet.h` if I remember it right.

PR #315, opened 2020-07-15
build: Check VS edition instead of assuming Professional

### LINKED ISSUE #308: Check which VS is installed instead of assuming "Professional"
Not everyone has VS Professional installed. I, for example, have Community only, but build scripts force the Professional directory.

We should check which one is installed before setting vcvars64 (or maybe ask the user to add it to path properly?).

JPTIZ:
Resolves #308.

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -1,19 +1,11 @@
 @echo off
 
 set __OLD_vcvars64=%vcvars64%
-set vcvars64="C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat"
 set CLICOLOR_FORCE=1
 
 set NINJAFLAGS=%*
 
-set vcvars64="C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat"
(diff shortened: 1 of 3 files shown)


JPTIZ:
[push-force](https://github.com/expertisesolutions/efl/compare/850018cfcb4147df66a8983a1c3213b79edc97ff..c54c198a06a3b5ae5123145dc97b03c07299a872): complemented the explanation for `setlocal EnableDelayedExpansion`.

JPTIZ:
[push-force](https://github.com/expertisesolutions/efl/compare/c54c198a06a3b5ae5123145dc97b03c07299a872..5eb76b04982800851125745f631408cd63405229): Fixed wrong calling of other batch script file. Calling another script should be done using the command `call`.

JPTIZ:
[push-force](https://github.com/expertisesolutions/efl/compare/5eb76b04982800851125745f631408cd63405229..1569961d01aee2bc74fceaa53324b7b99e4855bf): Fixed wrong usage of `%vcvars64%` (should be `!vcvars64!` because of that delayed expansion I mentioned).

JPTIZ:
Should be ok now.

PR #320, opened 2020-07-16
build: Properly handle --verbose arg

JPTIZ:
# Current scenario

In `configure.bat`, the user may want to add custom flags to meson calling:

```console
$ configure <extra args to meson>
```

Meanwhile, a "verbose" mode was also available for configure, which would send `-v` flag to the `clang-cl` compiler (making it always output its command line, so there would be no need to check every `.rsp` to see compiler flags).

# The problems

Previously, verbose output from configure only had effect if `--verbose` was the **first** argument. Not only that, we had a hidden bug on it:
- Since `%MESONFLAGS_EXTRA%` was used (instead of `!MESONFLAGS_EXTRA!`), no flags were sent to meson from `configure` command line (because of how delayed expansion works on Windows);
- By fixing the usage of  `!` instead of `%`, since there was no filter on extra args sent to meson, the `--verbose` flag was also being sent, which would make `meson` fail due to not having a `--verbose` flag.

# The fix
- Fix sending extra flags to meson replacing `%MESONFLAGS_EXTRA%` with `!MESONFLAGS_EXTRA!`;
- Filter `configure.bat`-specific args by not sending them to meson.

### DIFF
--- a/configure.bat
+++ b/configure.bat
@@ -1,14 +1,17 @@
 @echo off
 setlocal EnableDelayedExpansion
 
-if "%~1"=="--verbose" (
+:: Receive extra-flags
+set MESONFLAGS_EXTRA=%*
+
+:: Look for verbosity of this script and additional flags for meson
+(echo %MESONFLAGS_EXTRA% | findstr /i /c:"--verbose" >nul) && set VERBOSE=ON
+if defined VERBOSE (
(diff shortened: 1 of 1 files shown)


JPTIZ:
push-force: Added the PR description to commit message body.

PR #322, opened 2020-07-17
build: Adds -h,--help to configure.bat

JPTIZ:


JPTIZ:
> 
> 
> You build failed on windows CI.

Lol, indeed. Seems like one script is calling the wrong command or something, hm...Lemme see soon.

cauebs:
It's expecting some args that should be optional, and then it's justing printing the usage string and exiting with a successful error code, it seems.

JPTIZ:
FINALLY

PR #323, opened 2020-07-17
[Evil] Reimplementing basename

joaoantoniocardoso:
This PR aims to reimplement the missing `basename` in some way that it should work when using multiple threads.


src/lib/evil/evil_libgen.c
```diff
@@ -0,0 +1,27 @@
...
+{
+   // Expected from posix
+   if ((path == NULL) || (*path == '\0') ) return ".";
+
+   // Remove trailing '\\'
+   unsigned int len = strlen(path) -1;
+   if ((path[len] == '/') || (path[len] == '\\')) path[len] = '\0';
+
+   // Search for the last slash
+   while(--len) if ((path[len] == '/') || (path[len] == '\\')) break;
```
JPTIZ, line 22: I really prefer a more explicit structuring. There was even 3 commands on the same line.

I'm gonna give my try here:

```suggestion
   when (len > 0)
     {
        --len;
        if ((path[len] == '/') || (path[len] == '\\')) break;
     }
```

src/lib/evil/evil_libgen.c
```diff
@@ -0,0 +1,27 @@
...
+
+   if(!len) return path;
+   else return &path[len + 1];
```
JPTIZ, line 25: What `path` is `"a"`? Wouldn't it get a simple `\0`?

src/lib/evil/evil_libgen.h
```diff
@@ -0,0 +1,13 @@
+#ifndef __EVIL_LIBGEN_H__
+#define __EVIL_LIBGEN_H__
+
+#include <evil_api.h>
```
JPTIZ, line 4: You should not include `evil_api.h` except in `evil_private.h`. You can think that `evil_private.h` is "the lib", the others are just "organized parts of it".

So no one includes `evil_{something that is not 'private'}.h`, but `evil_private.h` includes everyone. Meanwhile no `evil_*` header includes any other `evil_*` (not even `evil_private.h`), every other file should include `evil_private.h` if it wants to use `evil`.

src/lib/evil/evil_libgen.c
joaoantoniocardoso: Thanks, fixed and added other tests.

src/lib/evil/evil_libgen.c
joaoantoniocardoso: Oh, I am quite used to this type of while structures and didn't thought it could be messy. See if it is more clear now. 

Despite my bare metal C background I prefer to read a decrementing while-if-break structure than a minimalist for loop, but maybe it would be more understandable with a for? I don't know what these compilers likes the most though.
```C
for(; (path[len] == '/') || (path[len] == '\\'); --len);
```
What do you think?

src/lib/evil/evil_libgen.h
joaoantoniocardoso: If I understood right, you are saying that we should be doing [this](https://github.com/expertisesolutions/efl/commit/c7389b77c9693c6af595ab55fae4e8381b22d675)?

Coquinho: yes, and no. In c7389b77c9693c6af595ab55fae4e8381b22d675 there are still some `evil_<thing>.h` that are including other `evil` headers besides `evil_private.h`. Accordingly to @JPTIZ they if they need anything from other `evil`, they should only include `evil_private.h`.

src/lib/evil/evil_libgen.h
```diff
@@ -0,0 +1,13 @@
+#ifndef __EVIL_LIBGEN_H__
+#define __EVIL_LIBGEN_H__
+
+#include <evil_private.h>
+
+#ifdef _MSC_VER
+# define basename evil_basename
+#endif
```
Coquinho, line 8: My question here is, should we guard this `evil` functions with `_MSC_VER` or with `HAVE_<func>`?

joaoantoniocardoso: I think `HAVE_<func>` makes more sense, but how to generate `HAVE_<func>`?

src/lib/evil/evil_libgen.h
joaoantoniocardoso: Nice, checkout 137a55a0ff4683c6420902c8e8934362d7454a99.

src/lib/evil/evil_libgen.h
joaoantoniocardoso: For now it seems ok to me doing the following:
```suggestion
#ifndef HAVE_BASENAME
#define HAVE_BASENAME
# define basename evil_basename
#endif
```
But in the case of using mingw, would meson define `HAVE_BASENAME` before us?

Coquinho: you have to include it to `header_checks/meson.build` like in #312 ([here](https://github.com/expertisesolutions/efl/pull/312/files#diff-2f0220d6614e01cb370018711bbe493a))

joaoantoniocardoso: Thanks, I'll do it.

Coquinho: for every `[foo, header]` on `function_checks` meson looks if there is any function `foo` at  `header` and if so it generated a `HAVE_FOO` with value 1.

joaoantoniocardoso: Ok, done. I kept the declaration outside of that guard because it is used in their test case.

src/lib/evil/evil_libgen.h
```diff
@@ -3,12 +3,19 @@
...
+
 #ifndef HAVE_DIRNAME
 # define HAVE_DIRNAME
 # define dirname evil_dirname
 #endif
 
-EVIL_API char* dirname (char* path);
+EVIL_API char* evil_basename (char* path);
+
+EVIL_API char* evil_dirname (char* path);
```
JPTIZ, line 18: Hm, you could've just alternate them, but ok haha

JPTIZ, line 18: Hm, you could've just alternate them, but ok haha

joaoantoniocardoso: yeah but it is fixing this too: `dirname` -> `evil_dirname`

Coquinho: I think he says something like:

```
#ifndef HAVE_BASENAME
# define HAVE_BASENAME
# define basename evil_basename
#endif

EVIL_API char* evil_basename (char* path);

#ifndef HAVE_DIRNAME
# define HAVE_DIRNAME
# define dirname evil_dirname
#endif

EVIL_API char* evil_dirname (char* path);
```

src/lib/evil/evil_private.h
```diff
@@ -40,7 +40,6 @@ extern "C" {
 # define _USE_MATH_DEFINES
 #endif
 
-#include "evil_basename.h"
```
Coquinho, line 43: Shouldn't is `evil_libgen.h` be included by `evil_private.h`?

src/lib/evil/meson.build
```diff
@@ -5,7 +5,6 @@ if target_machine.system() == 'windows'
   evil_src = []
 
   evil_src += files([
-    'evil_basename.c',
```
Coquinho, line 8: Shouldn't is `evil_libgen.c` be included evil source files?

src/lib/evil/evil_libgen.c
JPTIZ: I'll address this in a later patch.

joaoantoniocardoso:
@JPTIZ thanks, we can open another PR reworking basename plus dirname in future.

PR #324, opened 2020-07-21
build: Use installed openssl when available

JPTIZ:
Stops demanding OPENSSL_DIR environment variable in order to compile, using it only if defined.

### DIFF
--- a/configure.bat
+++ b/configure.bat
@@ -28,19 +28,15 @@ exit /B %errorlevel%
     )
     set envfile=
 
-    set all_set=1
-    if not defined OPENSSL_DIR set all_set=0
-
-    if %all_set%==1 (
-        @echo - Using OpenSSL: %OPENSSL_DIR%
-    ) else (
-        @echo At least one of the following variables were not set:
-        @echo     - OPENSSL_DIR: %OPENSSL_DIR%
-        @echo Please define them using by creating a "env.bat" file containing:
-        @echo     @set OPENSSL_DIR=^<your OpenSSL directory^>
-        exit /B 1
+    if not defined openssl_dir (
+        echo [INFO] "OPENSSL_DIR" environment variable is not defined. Relying
+        echo on meson to properly find OpenSSL with standard
+        echo `dependency('openssl'^)` command. If meson does not find OpenSSL,
+        echo consider setting it to your current OPENSSL library directory. In
+        echo this case, you may consider creating a env.bat file containing:
+        echo
+        echo     @set OPENSSL_DIR=^<your OpenSSL directory^>
     )
-    set all_set=
 
     set pkg_config_libdir=thistotallydoesnotexist
 exit /B 0
@@ -70,7 +66,6 @@ exit /B 0
 
     :: ------------------------------------------------------
     set MESONFLAGS=^
-            -Dopenssl_dir="%OPENSSL_DIR:"=%"^
             -Dnls=false^
             -Dsystemd=false^
             -Dgstreamer=false^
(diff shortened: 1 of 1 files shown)


configure.bat
```diff
@@ -28,20 +28,6 @@ exit /B %errorlevel%
...
-        @echo - Using OpenSSL: %OPENSSL_DIR%
-    ) else (
-        @echo At least one of the following variables were not set:
-        @echo     - OPENSSL_DIR: %OPENSSL_DIR%
-        @echo Please define them using by creating a "env.bat" file containing:
-        @echo     @set OPENSSL_DIR=^<your OpenSSL directory^>
-        exit /B 1
-    )
-    set all_set=
-
```
Coquinho, line 44: If we remove this completely, ones that don't have `openssl_dir` defined will never understand the problem.
We should maintain this as the `else` of `if defined openssl_dir`.

JPTIZ: But the "problem" does not exist anymore. We don't need to enforce `openssl_dir` by any means. If openssl is not going to be found for someone, meson will just not find it and then we'll no the problem. Letting this in a guide would be better than in the script, I think.

What I can do, tho, is...

JPTIZ: [this](https://github.com/expertisesolutions/efl/compare/890e774274e053c26dd821bb7f5941eb563bdcec..df233053e17d2a93ea39107fc73a777fdb90cad6).

configure.bat
```diff
@@ -28,19 +28,11 @@ exit /B %errorlevel%
...
-    ) else (
-        @echo At least one of the following variables were not set:
-        @echo     - OPENSSL_DIR: %OPENSSL_DIR%
-        @echo Please define them using by creating a "env.bat" file containing:
-        @echo     @set OPENSSL_DIR=^<your OpenSSL directory^>
-        exit /B 1
+    if not defined openssl_dir (
+        echo [INFO] "OPENSSL_DIR" environment variable is not defined. If meson
+        echo        does not find OpenSSL, consider setting it to your current
+        echo        OPENSSL library directory.
```
Coquinho, line 34: Why not maintain the original message and break? Actually, maybe changing it for something to make it clear that is only about OpenSSL like: 

```suggestion
        @echo OPENSSL_DIR variable was not set:
        @echo     - OPENSSL_DIR: %OPENSSL_DIR%
        @echo Please define it by creating a "env.bat" file containing:
        @echo     @set OPENSSL_DIR=^<your OpenSSL directory^>
        exit /B 1
```

JPTIZ: Because this makes it looks like it is obligatory. OpenSSL_DIR is not meant to be the default, it is optional.

JPTIZ: I mean, it is not even "looks like". It is literally mandating the user to define it.

configure.bat
```diff
@@ -28,19 +28,15 @@ exit /B %errorlevel%
...
-        @echo     @set OPENSSL_DIR=^<your OpenSSL directory^>
-        exit /B 1
+    if not defined openssl_dir (
+        echo [INFO] "OPENSSL_DIR" environment variable is not defined. Relying
+        echo on meson to properly find OpenSSL with standard
+        echo `dependency('openssl'^)` command. If meson does not find OpenSSL,
+        echo consider setting it to your current OPENSSL library directory. In
+        echo this case, you may consider creating a env.bat file containing:
+        echo
+        echo     @set OPENSSL_DIR=^<your OpenSSL directory^>
```
Coquinho, line 38: Much better, thanks.

Coquinho: those `^` should really be here?

JPTIZ: This is windows' `\` for escaping characters :D So yeah, they should. Explaining the excapings:

- `)` is considered a closing scope, `^)` is a literal ")";
- `<` and `>` file input/output redirs, `^<` and `^>` are literal lt and gt.

Coquinho: windows.....

PR #331, opened 2020-07-22
[win32] [examples] Adding guards to headers for windows compatibility

joaoantoniocardoso:
This is a rework of some commits necessary build examples on windows.
based on:
- 2a93955b05f83ab32d26749357fbb849f6c1c561
- f5778429c655e9641b60fb72557449171472927f
- 276d039c35d31df2f2a2adebad6240a373d27356

### DIFF
--- a/src/examples/ecore/ecore_audio_custom.c
+++ b/src/examples/ecore/ecore_audio_custom.c
@@ -2,11 +2,16 @@
 // gcc -o ecore_audio_custom ecore_audio_custom.c `pkg-config --libs --cflags ecore ecore-audio`
 
 #include <stdio.h>
-#include <sys/types.h>
-#include <sys/stat.h>
-#include <termios.h>
-#include <unistd.h>
-#include <fcntl.h>
+#ifndef _MSC_VER
+# include <sys/types.h>
+# include <sys/stat.h>
+# include <termios.h>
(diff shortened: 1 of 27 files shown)


src/examples/ecore/ecore_audio_custom.c
```diff
@@ -2,11 +2,15 @@
...
-#include <fcntl.h>
+#ifndef _MSC_VER
+# include <sys/types.h>
+# include <sys/stat.h>
+# include <termios.h>
+# include <unistd.h>
+# include <fcntl.h>
+#else
+# include <evil_private.h>
+#endif
```
Coquinho, line 13: shouldn't this be like:
```suggestion
#endif

#ifdef _WIN32
# include <evil_private.h>
#endif
```
?

joaoantoniocardoso: You're right

src/examples/ecore/ecore_evas_extn_plug_example.c
```diff
@@ -16,7 +16,6 @@
 
 #include <Ecore.h>
 #include <Ecore_Evas.h>
-#include <unistd.h>
```
Coquinho, line 19: We don't need `evil_private.h`?

Issue #190, opened 2020-06-25
Evil tests are slow, so maybe evil functions are slow


cauebs:
We had to increase the time limit on the evil test suite so it wouldn't timeout. I checked and it's not getting stuck a single spot, it's really just slow all the way through. Functions like `getenv`, `setenv` and `realpath`, for example, appear to be taking too long to run.

It builds fine and tests still pass, so I'm not marking this as a bug, but it doesn't look right to me...

vtorri:
strange. I can understand that setenv and realpath are taking time, but getenv is a function provided by Windows, so...

did you do some profiling ? either with Visual Studio or very sleepy ?

PR #335, opened 2020-07-22
evil: Remove unnecessary delay for evil startup

Coquinho:
Without this `ecore-suite`, `efl-app` and `evas-suite` gets timeout:
```
25/29 ecore-suite               TIMEOUT        30.12s
26/29 efl-app                   TIMEOUT        30.14s
...
29/29 evas-suite                TIMEOUT        60.07s
```

As a side effect, this speed up `evil`'s tests thus closing #190.

Based on 8d083963454a9da3045a53315954bd321c75230e

### DIFF
--- a/src/lib/evil/evil_main.c
+++ b/src/lib/evil/evil_main.c
@@ -41,24 +41,10 @@ evil_init(void)
 
    if (!QueryPerformanceFrequency(&freq))
        return 0;
-
    _evil_time_freq = freq.QuadPart;
 
-   /* be sure that second + 1 != 0 */
-   while (second == 59)
-     {
-        GetSystemTime(&st);
-        second = st.wSecond;
-     }
-
-   /* retrieve the tick corresponding to the time we retrieve above */
-   while (1)
-     {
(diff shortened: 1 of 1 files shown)


cauebs:
Wow, this closes #190! Lowers test time from 33s to 2s on my machine. As a neat side effect, it will trim these 30s from the CI runs as well! :rocket:

src/lib/evil/evil_main.c
```diff
@@ -44,21 +44,8 @@ evil_init(void)
...
-     }
-
-   /* retrieve the tick corresponding to the time we retrieve above */
-   while (1)
-     {
-        GetSystemTime(&st);
-        QueryPerformanceCounter(&count);
-        if (st.wSecond == second + 1)
-          break;
-     }
```
cauebs, line 61: @felipealmeida could you shed some light on what this did and why it's no longer needed?

Coquinho: Apparently this was used for an old implementation of `gettimeofday` that was abandoned. And this change is safe.

PR #339, opened 2020-07-24
eina: Check if thread is null before accessing its values on eina_thread_cancel

Coquinho:
Based on: ae462147fd6beff0ad3c0a96280e5e31ba3d1f0a

### DIFF
--- a/src/lib/eina/eina_thread_win32.c
+++ b/src/lib/eina/eina_thread_win32.c
@@ -222,14 +222,16 @@ eina_thread_cancel(Eina_Thread t)
     Eina_Bool ret = EINA_FALSE;
     Thread_t *thr = (Thread_t *) t;
 
-    EnterCriticalSection(&thr->cancel_lock);
-    if (thr->cancellable)
+    if (thr)
       {
-          thr->cancel = EINA_TRUE;
-          ret = EINA_TRUE;
+         EnterCriticalSection(&thr->cancel_lock);
+         if (thr->cancellable)
+           {
+              thr->cancel = EINA_TRUE;
+              ret = EINA_TRUE;
+           }
+         LeaveCriticalSection(&thr->cancel_lock);
       }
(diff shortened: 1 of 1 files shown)


cauebs:
Shouldn't this be checked at the call site instead? Shouldn't the function presume that the parameters are valid? Honest question, because I'm not sure. If we go this way, this check should be added to all similar functions.

Coquinho:
> Shouldn't this be checked at the call site instead? Shouldn't the function presume that the parameters are valid? Honest question, because I'm not sure. If we go this way, this check should be added to all similar functions.

Having a look at [[eina_thread_cancel](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/src/lib/eina/eina_thread_posix.c#L270) from posix, this is done at the function too, so maybe it should always be done at the function. 

Either way, maybe we should open an issue to discuss it?

cauebs:
Well, I can see that in both win32 and posix implementations `eina_thread_set_name` doesn't check if `t` is valid, so I really can't tell if there even is a convention being followed... I don't oppose to this change, but I really think everyone should think more carefully about APIs, contracts and conventions (not pointing fingers to anyone in particular, just a general worry).

Coquinho:
> Well, I can see that in both win32 and posix implementations `eina_thread_set_name` doesn't check if `t` is valid, so I really can't tell if there even is a convention being followed... I don't oppose to this change, but I really think everyone should think more carefully about APIs, contracts and conventions (not pointing fingers to anyone in particular, just a general worry).

Sure sure. What we do as it seems that there is no convention?

cauebs:
This PR doesn't raise any serious issues, so I guess we just merge it as is.

Coquinho:
> This PR doesn't raise any serious issues, so I guess we just merge it as is.

okidoki, if you want to discuss this further, we may open an issue, or even bring this to `#edevelop`

PR #340, opened 2020-07-24
elm: Remove declaration from the wrong library

Coquinho:
`_efl_startup_time` is declared in [Ecore library](https://github.com/expertisesolutions/efl/blob/devs/coquinho/clean/evas-fixes/src/lib/ecore/ecore.c#L47).
It shouldn't be declared in Elementary because it causes mismatches with
importing and exporting of symbol by MSVC's linker.

Original: 06552189d14ff466f93345c87ecdcbf0b7ca526f

### DIFF
--- a/src/lib/elementary/Efl_Ui.h
+++ b/src/lib/elementary/Efl_Ui.h
@@ -74,8 +74,6 @@
 extern "C" {
 #endif
 
-ELM_API extern double _efl_startup_time;
-
 /** Successfully applied the requested style from the current theme. */
 ELM_API extern Eina_Error EFL_UI_THEME_APPLY_ERROR_NONE;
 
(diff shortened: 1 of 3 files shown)


cauebs:
But then aren't files like `src/lib/elementary/elm_main.c ` and `src/tests/elementary/suite_helpers.c` missing an include? They're probably getting that definition indirectly via other includes, but I understand that is not a good practice.

Coquinho:
> src/lib/elementary/elm_main.c

I think we can make this explicitly, but then we are no going o the minor effort, what do you guys think?

JPTIZ:
> 
> 
> > src/lib/elementary/elm_main.c
> 
> I think we can make this explicitly, but then we are no going o the minor effort, what do you guys think?

If this definition is on ecore, then we need ecore, then makes sense to make it explicit here. Wouldn't bother to not approve now ;D

PR #341, opened 2020-07-24
evil: Implement evil_localtime_r using localtime

Coquinho:
Implemente `evil_localtime_r` using win32 `localtime` function.  If
`HAVE_lOCALTIME_R` is not defined, do it and define `localtime_r` as
`evil_localtime_r`.

Original: 0832953c9eae40851f35b300c25f5a19e35879ef

### DIFF
--- a/src/lib/evil/evil_time.c
+++ b/src/lib/evil/evil_time.c
@@ -11,11 +11,8 @@
 EVIL_API struct tm *
 evil_localtime_r(const time_t * time, struct tm * result)
 {
-  *result = *localtime(time);
-  if (!errno)
-    return result;
-  else
-    return NULL;
+   *result = *localtime(time);
+   return result;
 }
 
 /*
--- a/src/lib/evil/evil_time.c
+++ b/src/lib/evil/evil_time.c
@@ -11,8 +11,14 @@
 EVIL_API struct tm *
 evil_localtime_r(const time_t * time, struct tm * result)
 {
-   *result = *localtime(time);
-   return result;
+   struct tm* _result = localtime(time);
+
+   if (_result)
+     {
+        *result = *_result;
+        return result;
+     }
+   return NULL;
 }
 
 /*
--- a/src/lib/evil/evil_time.c
+++ b/src/lib/evil/evil_time.c
@@ -8,8 +8,13 @@
 
 #include "evil_private.h"
 
-inline struct tm *localtime_r(const time_t * time, struct tm * result)
+EVIL_API struct tm *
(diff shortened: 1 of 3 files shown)


src/lib/evil/evil_time.c
```diff
@@ -8,8 +8,13 @@
...
 #include "evil_private.h"
 
-inline struct tm *localtime_r(const time_t * time, struct tm * result)
+EVIL_API struct tm *
+evil_localtime_r(const time_t * time, struct tm * result)
 {
+  *result = *localtime(time);
+  if (!errno)
+    return result;
+  else
```
JPTIZ, line 17: No need for an else, actually.

src/lib/evil/evil_time.c
```diff
@@ -8,9 +8,11 @@
 
 #include "evil_private.h"
 
-inline struct tm *localtime_r(const time_t * time, struct tm * result)
+EVIL_API struct tm *
+evil_localtime_r(const time_t * time, struct tm * result)
 {
-    return NULL;
+   *result = *localtime(time);
+   return result;
```
JPTIZ, line 15: Hmmm wait wait, actually have another problem here. I think you'll have to first store it in a temporary `struct tm*`, because if `localtime` returns `NULL`, `*localtime` will try to derreference a `NULL`.

Maybe this?

```suggestion
   struct tm* _result = localtime(time);
   
   if (_result)
     {
        *result = *_result;
        return result; // or should we return `_result`? Are they the same pointer? How to deal?
     }
   return NULL;
```

Coquinho: ok...so why not the original commit: 73e3740ca29a7983d7bccb5a30564dc9e4c83e60?

JPTIZ: ....because it doesn't solve the problem.

Coquinho: nice, could you explain?

JPTIZ: I already explained. It has the very same issue I commented, look closely to this line:

```c
*result = *localtime(time);
```

Now re-read:

> Hmmm wait wait, actually have another problem here. I think you'll have to first store it in a temporary `struct tm*`, because **if `localtime` returns `NULL`, `*localtime` will try to derreference a `NULL`.**

JPTIZ: That commit still have that very same line

Coquinho: ok, that makes sense, thanks

Issue #307, opened 2020-07-14
Use vcpkg to install dependencies


JPTIZ:
- **Coming soon**

PR #346, opened 2020-07-28
build, ci: Support custom CMake modules, add vcpkg to CI

JPTIZ:
This makes our EFL changes less dependant on wraps and a little more flexible making usage of `vcpkg` for dependency management and meson's `cmake_args` to send extra arguments to cmake when looking for package.

Specifically, this PR:
- Installs some of the available libraries in `vcpkg`;
- Fixes cache issues with Github CI by upgrading from `cache@v1` to `cache@v2`;
- Upgrades version restriction in lua version to consider Lua 5.3 as well;
- Slightly enhances "Lua not found" error messages (stating which names it tried).

**Issues**

Unfortunately, it wasn't possible to get rid of the following wrap-dependencies:
- getopt, because meson is not finding it from installed vcpkg package (and I also there's no `FindGetopt.cmake` in `Modules` dir);
- lua, because even if it finds the dependency, there's a linkage problem (with `luaL_openlib` not found, as can be seen [in a previous CI build log](https://github.com/expertisesolutions/efl/runs/955863806?check_suite_focus=true#step:31:4053));
- luajit, because cmake can't find it for some reason (probably because there's no `FindLuajit.cmake`?);
- pcre, by the same reason of getopt.

I'm letting lua and luajit installed on CI for cache reasons, since we'll be tackling this later.


JPTIZ:
[push-force](https://github.com/expertisesolutions/efl/compare/2a01192e1af64f26877228b306decd25dac1b66d..b5cd569369c3875623e7d7767d3f59ad6ed13af6): Add libjpeg (using jpeg-turbo backend), lua and zlib.

JPTIZ:
[push-force](https://github.com/expertisesolutions/efl/compare/790de97e4192657ba66a60ccc31752bf6b9b20e8..90e882e513a15f3824d6749b50aeaa5065098ef4): Add check and getopt.

Coquinho:
Should we send to native-windows what is already working?

JPTIZ:
> 
> 
> Should we send to native-windows what is already working?

The `ector-suite` test crashes for unknown reasons. I'm trying to make #367 work in order to debug here.

JPTIZ:
FINALLY²

It was a path issue the whole moment. Had to add vcpkg's installed .dll dirs (for both debug and release dlls). And just to make sure I added the internal libs' paths too.

Coquinho:
@JPTIZ CI still doesn't work :/

JPTIZ:
> 
> 
> @JPTIZ CI still doesn't work :/

Oh that's because I tried to play smart-boy in freetype and jpeg dependency fetching :^D Gonna quickly solve that

JPTIZ:
> 
> 
> @JPTIZ CI still doesn't work :/

*clears throat*

IT IS ALIVE (again)

meson_options.txt
```diff
@@ -381,3 +381,13 @@ option('efl-one',
...
+
+option('cmake_args',
+  type: 'string',
+  value : '',
+  description : '')
+
+option('cmake_module_path',
+  type: 'string',
+  value : '',
+  description : '')
```
joaoantoniocardoso, line 393: Could you provide some description?

JPTIZ: Oh, true, yes I can.

.github/workflows/windows.yml
```diff
@@ -10,38 +10,44 @@ jobs:
...
     - uses: actions/checkout@v2
-    - name: Cache Nasm install
-      uses: actions/cache@v1
-      id: cache-nasm
-      env:
-        cache-name: cache-nasm
-      with:
-        path: "C:\\Program Files\\Nasm"
-        key: cache-nasm-v1
-        restore-keys: cache-nasm-v1
```
Coquinho, line 21: `NASM` is already on `OpenSSL`?

.github/workflows/windows.yml
```diff
@@ -10,38 +10,44 @@ jobs:
...
     - name: Cache LLVM install
-      uses: actions/cache@v1
+      uses: actions/cache@v2
```
Coquinho, line 14: Didn't get the point of this change. Could you explain to me?

.github/workflows/windows.yml
```diff
@@ -10,38 +10,44 @@ jobs:
...
+    - name: Show vcpkg cache info
+      env:
+        VCPKG_INFO: ${{ toJSON(steps.cache-vcpkg) }}
+      run: echo "${VCPKG_INFO}"
+    - name: Has hit cache for vcpkg?
+      if: steps.cache-vcpkg.outputs.cache-hit == 'true'
+      run: echo Hit!
+    - name: Has missed cache for vcpkg?
+      if: steps.cache-vcpkg.outputs.cache-hit != 'true'
+      run: echo Missed!
```
Coquinho, line 43: This seems as a debugging thing or am I wrong?

.github/workflows/windows.yml
```diff
@@ -55,35 +61,47 @@ jobs:
...
+      run: git clone https://github.com/microsoft/vcpkg
+    - name: Checkout to f4bd6423
+      if: steps.cache-vcpkg.outputs.cache-hit != 'true'
+      run: git checkout f4bd6423
+      working-directory: .\vcpkg
+    - name: Run vcpkg bootstrap
+      if: steps.cache-vcpkg.outputs.cache-hit != 'true'
+      run: .\vcpkg\bootstrap-vcpkg.bat
+    - name: Install openssl (vcpkg)
+      if: steps.cache-vcpkg.outputs.cache-hit != 'true'
```
Coquinho, line 83: Are those `if` still necessary?

.github/workflows/windows.yml
```diff
@@ -119,6 +159,42 @@ jobs:
...
+        dir "${{ github.workspace }}\build\src\lib\eet"
+        dir "${{ github.workspace }}\build\src\lib\efl"
+        dir "${{ github.workspace }}\build\src\lib\eina"
+        dir "${{ github.workspace }}\build\src\lib\emile"
+        dir "${{ github.workspace }}\build\src\lib\eo"
+        dir "${{ github.workspace }}\build\src\lib\evil"
+        dir "${{ github.workspace }}\build\subprojects\getopt"
+        dir "${{ github.workspace }}\build\subprojects\pcre-8.37"
+        dir "${{ github.workspace }}\build\src\lib\ector"
+    - name: Check path
```
Coquinho, line 196: So for now on, for each lib we enable we need to add its dll here?

Coquinho: Could you sort those, please? =D

JPTIZ: Yes (actually let me do a quick test before that) and yes.

Coquinho:
Will this close #307?

.github/workflows/windows.yml
JPTIZ: Yes. We just want to run bootstrap/etc. if the cache hits (you're asking if it shouldn't automatic ignored by the CI?).

Coquinho: So shouldn't all package installations with `vcpkg` have that? Or the point is to don't cache `freetype`, `llibjpeg`, `lua`, `zlib`, and `getopt` and only cache `openssl`?

.github/workflows/windows.yml
Coquinho: `check` dll is don't need to be set like this too?

Coquinho: oh, our dependencies don't need that it seems.

JPTIZ:
> 
> 
> `luajit` will remain as a meson wrap?
> `lua` and `getopt` wrap weren't removed

But they are removed for me o.o Where exactly did you see?

JPTIZ:
Oh I see, I didn't rebase it, so I couldn't see them. Should be fixed now.

.github/workflows/windows.yml
JPTIZ: Yup, they can be removed, to the first answer changes to "Nope" :D

JPTIZ: (For reference: https://github.com/expertisesolutions/efl/runs/950753122?check_suite_focus=true#step:20:1)

meson_options.txt
```diff
@@ -387,3 +387,8 @@ option('efl-one',
   value : false,
   description : 'Build special set of .so´s as a single .so'
 )
+
+option('cmake_args',
+  type: 'string',
+  value : '',
+  description : 'Extra arguments for cmake when searching for dependencies.')
```
Coquinho, line 394: I think it is a good idea to stay in the pattern:
```suggestion

option('cmake_args',
  type: 'string',
  value : '',
  description : 'Extra arguments for cmake when searching for dependencies.'
)
```

JPTIZ: Done.

.github/workflows/windows.yml
JPTIZ: Uuuuuuuuuuuuuuuuh yes, I think. Let me see how CI behaves on that.

.github/workflows/windows.yml
JPTIZ: 1. Why _not_ using v2, since it is the updated? :D
2. I explained in Slack, but to clarify here too: v1 doesn't cache properly. And worse, when it caches, not everytime it sets `cache-hit=true`. That's why some things were executing everytime even if there was an `if steps.<something>.cache-hit == 'true'`.

.github/workflows/windows.yml
```diff
@@ -117,6 +157,26 @@ jobs:
...
+    - name: Add build libs to path
+      run: |
+        echo "::add-path::${{ github.workspace }}\build\subprojects\getopt"
+        echo "::add-path::${{ github.workspace }}\build\subprojects\pcre-8.37"
+        echo "::add-path::${{ github.workspace }}\build\src\lib\ector"
+    - name: Check built EFL libs
+      run: |
+        dir "${{ github.workspace }}\build\subprojects\getopt"
+        dir "${{ github.workspace }}\build\subprojects\pcre-8.37"
+        dir "${{ github.workspace }}\build\src\lib\ector"
```
Coquinho, line 177: What about the others efl dll's?

.github/workflows/windows.yml
Coquinho: hummm, but ector is still needed?

.github/workflows/windows.yml
JPTIZ: I have no idea if is it or not, but at least vcpkg's openssl doesn't require to priorly explicitly install nasm.

.github/workflows/windows.yml
Coquinho: lol, okay, thanks

.github/workflows/windows.yml
JPTIZ: It _probably_ installs internally (at least it installs strawberry-perl), but...

meson_options.txt
JPTIZ: Done.

.github/workflows/windows.yml
JPTIZ: I was thinking about letting it there at first, but to be honest...it is too situational to keep it there forever.

JPTIZ: Done.

.github/workflows/windows.yml
Coquinho: haha, okay then, if it works :D

.github/workflows/windows.yml
```diff
@@ -82,9 +82,9 @@ jobs:
     - name: Install libjpeg (vcpkg)
       if: steps.cache-vcpkg.outputs.cache-hit != 'true'
       run: .\vcpkg\vcpkg install libjpeg-turbo:x64-windows
-    - name: Install lua (vcpkg)
+    - name: Install luajit (vcpkg)
       if: steps.cache-vcpkg.outputs.cache-hit != 'true'
-      run: .\vcpkg\vcpkg install lua:x64-windows
+      run: .\vcpkg\vcpkg install luajit:x64-windows
```
Coquinho: Why not both? At native-windows we have both with meson wraps

JPTIZ: "Why not both" is a great question.

JPTIZ: Done.

.github/workflows/windows.yml
JPTIZ: Didn't see ector there. Should be fixed now.

.github/workflows/windows.yml
JPTIZ: Ok, CI keeps asking vcpkg to install them (although it just ends with a "but it is already installed!"). Now I'm caching everyone.

Coquinho:
It's broken again.

.github/workflows/windows.yml
Coquinho: oh, right it was fixed. So why dll's were needed before and now they are not?

.github/workflows/windows.yml
```diff
@@ -55,35 +55,55 @@ jobs:
...
+      if: steps.cache-vcpkg.outputs.cache-hit != 'true'
+      run: .\vcpkg\vcpkg install freetype:x64-windows
+    - name: Install libjpeg (vcpkg)
+      if: steps.cache-vcpkg.outputs.cache-hit != 'true'
+      run: .\vcpkg\vcpkg install libjpeg-turbo:x64-windows
+    - name: Install lua (vcpkg)
+      if: steps.cache-vcpkg.outputs.cache-hit != 'true'
+      run: .\vcpkg\vcpkg install lua:x64-windows
+    - name: Install luajit (vcpkg)
+      run: .\vcpkg\vcpkg install luajit:x64-windows
```
Coquinho, line 89: I think this way luajit will not be cached, right?

JPTIZ: It will, but I had to ensure it would run at least once.

JPTIZ:
> 
> 
> Nice work, just a question: luajit/lua is still using wrap?

I edited this PR's description for more clarification.

PR #347, opened 2020-07-28
[Evas] Fix fonts for win32

joaoantoniocardoso:
originally:  
  - ad327702366ba2939d952f6d7718d2d1c6b91d8b
  - 681218b9fe3824bf0d817531b7abfe7a1277f4bd

### DIFF
--- a/src/examples/evas/evas-text.c
+++ b/src/examples/evas/evas-text.c
@@ -316,7 +316,13 @@ WinMain(HINSTANCE hInstance EINA_UNUSED
     * example */
    struct text_preset_data init_data =
    {
+#if !defined(_MSC_VER) || defined(HAVE_FONTCONFIG)
       .font = {"DejaVu", "Courier", "Utopia"},
+#else
+      .font = {"C:\\Windows\\Fonts\\arial.ttf",
+               "C:\\Windows\\Fonts\\cour.ttf",
+               "C:\\Windows\\Fonts\\times.ttf"},
+#endif
       .text = {BLACK, WHITE, GREY, RED, GREEN, BLUE},
       .shadow = {WHITE, BLUE, GREEN, RED},
       .outline = {WHITE, RED, GREEN, BLUE},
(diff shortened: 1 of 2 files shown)


src/lib/evas/canvas/evas_font_dir.c
```diff
@@ -603,6 +603,10 @@ evas_font_name_parse(Evas_Font_Description *fdesc, const char *name)
    const char *end;
 
    end = strchr(name, ':');
+#ifdef _WIN32
```
Coquinho, line 606: Shouldn't this be `_MSC_VER`?

src/examples/evas/evas-text.c
```diff
@@ -316,7 +316,11 @@ WinMain(HINSTANCE hInstance EINA_UNUSED
     * example */
    struct text_preset_data init_data =
    {
+#if !defined(_WIN32) || defined(HAVE_FONTCONFIG)
```
Coquinho, line 319: Shouldn't this be `_MSC_VER`?

src/examples/evas/evas-text.c
```diff
@@ -316,7 +316,11 @@ WinMain(HINSTANCE hInstance EINA_UNUSED
     * example */
    struct text_preset_data init_data =
    {
+#if !defined(_MSC_VER) || defined(HAVE_FONTCONFIG)
       .font = {"DejaVu", "Courier", "Utopia"},
+#else
+      .font = {"C:\\Windows\\Fonts\\Arial.ttf", "C:\\Windows\\Fonts\\Cour.ttf", "C:\\Windows\\Fonts\\Comic.ttf"},
```
JPTIZ, line 322: Comic Sans? Really?

Well, I think we should get similar font alternatives for each one. Courier with Courier New seems fine, but Utopia & Comic veeery very different (one is serif, the other is sans). For replacements I would recommend:

- DejaVu ->Arial (for shape) or Verdana (for spacings)
- Utopia -> Times New Roman (see https://www.fontshop.com/families/utopia/similar).

PR #351, opened 2020-07-28
clang: Fix enumeration signed overflow

joaoantoniocardoso:
originally: a0562d5dd72c3d0cb83f8ab6f429707fa37a340e

### DIFF
--- a/src/lib/edje/Edje_Common.h
+++ b/src/lib/edje/Edje_Common.h
@@ -1894,7 +1894,11 @@ typedef enum _Edje_Tween_Mode
    EDJE_TWEEN_MODE_CUBIC_BEZIER      = 11,       /**< Cubic Bezier tween mode value */
    EDJE_TWEEN_MODE_LAST              = 12,       /**< Last tween mode value */
    EDJE_TWEEN_MODE_MASK              = 0xff,     /**< Mask tween mode value */
+#ifdef _MSC_VER
+   EDJE_TWEEN_MODE_OPT_FROM_CURRENT  = INT_MAX /**< Options from current tween mode value */
+#else
    EDJE_TWEEN_MODE_OPT_FROM_CURRENT  = (1 << 31) /**< Options from current tween mode value */
+#endif
 } Edje_Tween_Mode;
 
 /**


src/lib/edje/Edje_Common.h
```diff
@@ -1894,7 +1894,11 @@ typedef enum _Edje_Tween_Mode
    EDJE_TWEEN_MODE_CUBIC_BEZIER      = 11,       /**< Cubic Bezier tween mode value */
    EDJE_TWEEN_MODE_LAST              = 12,       /**< Last tween mode value */
    EDJE_TWEEN_MODE_MASK              = 0xff,     /**< Mask tween mode value */
+#ifdef _MSC_VER
+   EDJE_TWEEN_MODE_OPT_FROM_CURRENT  = INT_MAX /**< Options from current tween mode value */
```
Coquinho, line 1898: I didn't get why this is the way to solve it, could you explain to me? some ELI5 haha

joaoantoniocardoso:
Question here: `(1 << 31)` is `2147483648`, while `INT_MAX` is `2147483647`. shouldn't be INT_MAX for all platforms?

joaoantoniocardoso:
> [e7491a7](https://github.com/expertisesolutions/efl/commit/e7491a776c831828af776f30898d35463a29ed7f) is about enum overflow too, maybe is a good idea to join them?

I don't think so, both seems to be unrelated to each other.

felipealmeida:
> 
> 
> > [e7491a7](https://github.com/expertisesolutions/efl/commit/e7491a776c831828af776f30898d35463a29ed7f) is about enum overflow too, maybe is a good idea to join them?
> 
> I don't think so, both seems to be unrelated to each other.
> 
> Question here: `(1 << 31)` is `2147483648`, while `INT_MAX` is `2147483647`. shouldn't be INT_MAX for all platforms?

You can't guarantee this value is not used by third-party. This would be a ABI-breaking-change. Besides, enums can be unsigned in some platforms (that's why they do not break as it does on Windows) which avoids the overflow

src/lib/edje/Edje_Common.h
joaoantoniocardoso: An enumeration has a size (at least enough to represent an integer) and so they cannot represent values out of its range. For example if in some architecture the integer holds 32 bits, so its maximum value will be 2^31-1, and `(1<<31)` is `2^32`.

PR #352, opened 2020-07-28
ci: Minimize logs noise.

JPTIZ:
Meson has a flag to show only failed tests' logs (see [other test options section of manual](https://mesonbuild.com/Unit-tests.html#other-test-options), so this will make test loggin a lot less noisy.
But to not loose potential debugging info, the logs are still shown after execution, but in another task.

I also took the chance to rename some steps in order to make them more..."fitting".


cauebs:
@cauebs approved these changes

PR #353, opened 2020-07-28
Preserve line endings when checking out eolian_gen reference files

cauebs:
We are currently setting `core.autocrlf` to `input` on Windows, to checkout files without converting line endings to CRLF, but that means we expect anyone cloning the repo to do the same or tests will fail.

Now, I'm setting the line ending for the specific files that **need** to use LF.

Credits to @JPTIZ for the finding!

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -96,8 +96,6 @@ jobs:
     runs-on: windows-latest
     needs: build
     steps:
-    - name: set git line ending config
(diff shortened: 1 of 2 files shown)


cauebs:
(I reverted the change to the CI because it's no longer necessary)

JPTIZ:
Maybe we should also open an issue to change the current implementation to use the proper line-ending, what do you think?

PR #342, opened 2020-07-24
[Deps] Elua

joaoantoniocardoso:
Elua is building when using luajit.

Test result:
`27/27 elua-suite                FAIL           5.13s (exit status 3221225477 or signal 3221225349 SIGinvalid)`

Depends on #343 .

PR #356, opened 2020-07-28
build: disable elua by default, plus nicer detection

joaoantoniocardoso:
> Elua is now disabled by default. There are some other changes:
> 
> 1) Elua scripts are only installed if Elua is enabled
> 2) Lua bindings are only installed if Elua is enabled
> 3) Elua with interpreter is clearly experimental and will message

Originally: c3a1060b94ae9df82e8406b481e8bd5fe5741df5

This PR conflicts with #342 .

### DIFF
--- a/src/bindings/meson.build
+++ b/src/bindings/meson.build
@@ -18,6 +18,10 @@ if (bindings.contains('cxx') == false and bindings.contains('mono'))
   )
 endif
 
+if bindings.contains('lua') and not have_elua
+  error('Elua is necessary for Lua bindings')
+endif
+
 foreach binding : bindings_order
   if bindings.contains(binding)
     subdir(join_paths( binding))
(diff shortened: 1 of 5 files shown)


src/bindings/meson.build
```diff
@@ -18,6 +18,10 @@ if (bindings.contains('cxx') == false and bindings.contains('mono'))
   )
 endif
 
+if bindings.contains('lua') and not have_elua
```
JPTIZ, line 21: Where does this `have_elua` comes from?

joaoantoniocardoso: [here](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/meson.build#L383v) you are, sir

JPTIZ: Lovely :D Thanks

PR #357, opened 2020-07-29
win32: Fix eina modules search with meson install

joaoantoniocardoso:
`EINA_PATH_SEP_[CS]` are special **C**haracter and **S**tring definitions with the correct path separator for the system (`\\` for `_WIN32`).
This PR changes all Evas module path related functions to use it.

Originally 426386578bb6f4e045474dbf1d94a502f29ff18b


joaoantoniocardoso:
Closing as it is not needed now.

PR #358, opened 2020-07-29
win32: fix multiple rules generate for elementary test

joaoantoniocardoso:
The Elementary build on windows stops on the following error:

```
ninja: error: build.ninja:10612: multiple rules generate src/bin/elementary/elementary_test.pdb [-w dupbuild=err]
```

There is no need for library elementary_test_lib_name at all, so we are removing it with the upstream commit 338a529af190f0c9c285e932a765f39ae4047a03.

Originally c74e968991c663f23f67118050630b8c4e684670 (from evas-fixes)
Solved with: 338a529af190f0c9c285e932a765f39ae4047a03

### DIFF
--- a/src/bin/elementary/meson.build
+++ b/src/bin/elementary/meson.build
@@ -169,11 +169,9 @@ elementary_test_src = [
 
 if sys_windows == false
   link_args = ['-rdynamic', '-fPIC', '-pie']
-  link_args_lib = ['-rdynamic', '-fPIC']
   package_c_args =  package_c_args + ['-fPIC']
 else
   link_args = []
-  link_args_lib = []
 endif
 elementary_test = executable('elementary_test',
         elementary_test_src,
@@ -185,18 +183,6 @@ elementary_test = executable('elementary_test',
           '-Delementary_test_DATA_DIR="'+join_paths(dir_data,'elementary')+'"'
         ],
         link_args: link_args
-)
-
-library('elementary_test',
-        elementary_test_src,
-        dependencies: [elementary, intl] + elementary_deps + elementary_pub_deps,
-        install: false,
-        c_args : package_c_args + [
-          '-Delementary_test_BIN_DIR="'+dir_bin+'"',
-          '-Delementary_test_LIB_DIR="'+dir_lib+'"',
-          '-Delementary_test_DATA_DIR="'+join_paths(dir_data,'elementary')+'"'
-        ],
-        link_args: link_args_lib
 )
 
 elementary_config_src = [


src/bin/elementary/meson.build
```diff
@@ -187,7 +187,11 @@ elementary_test = executable('elementary_test',
         link_args: link_args
 )
 
-library('elementary_test',
+elementary_test_lib_name = 'elementary_test'
+if sys_windows
+elementary_test_lib_name = 'elementary_test_lib'
+endif
```
Coquinho, line 193: please add some indentation here.

JPTIZ:
This is doubly weird.

First because meson shouldn't create clashing names if the targets aren't even of the same "type" (one is an executable, the other is a library).

Second because they aren't even built in the same directory.

JPTIZ:
> 
> 
> This is doubly weird.
> 
> First because meson shouldn't create clashing names if the targets aren't even of the same "type" (one is an executable, the other is a library).
> 
> Second because they aren't even built in the same directory.

Oh actually I confused things here: this is in `src/bin/elementary`, not `src/lib/elementary`.

But to be honest: why is even a library here? Seems like some artifact that should be thrown away.

src/bin/elementary/meson.build
```diff
@@ -187,7 +187,11 @@ elementary_test = executable('elementary_test',
         link_args: link_args
 )
 
-library('elementary_test',
+elementary_test_lib_name = 'elementary_test'
+if sys_windows
+  elementary_test_lib_name = 'elementary_test_lib'
+endif
+library(elementary_test_lib_name,
```
Coquinho, line 194: Wait...is this even used somewhere?
Shouldn't it be something like: `<lib_var> = library(...)`, so we could use it as `<lib_var>`?

Coquinho:
> I suggest that this PR is abbandoned and be replaced with a new one, since this doesn't even exist on upstream: http://git.enlightenment.org/core/efl.git/tree/src/bin/elementary/meson.build#n179
> 
> As it can be seen by that line, it should actually be an executable, which is just above this `library` in our branches. The new PR should remove this declaration entirely.

It was [removed](https://github.com/Enlightenment/efl/commit/338a529af190f0c9c285e932a765f39ae4047a03#diff-9c0e9ba1b6c2a0ee64c2e756ad729690) this month XD

@joaoantoniocardoso could you try applying that commit from upstream to see if it solves your problem? If it does we could change this PR to apply it.

joaoantoniocardoso:
> > I suggest that this PR is abbandoned and be replaced with a new one, since this doesn't even exist on upstream: http://git.enlightenment.org/core/efl.git/tree/src/bin/elementary/meson.build#n179
> > As it can be seen by that line, it should actually be an executable, which is just above this `library` in our branches. The new PR should remove this declaration entirely.
> 
> It was [removed](https://github.com/Enlightenment/efl/commit/338a529af190f0c9c285e932a765f39ae4047a03#diff-9c0e9ba1b6c2a0ee64c2e756ad729690) this month XD
> 
> @joaoantoniocardoso could you try applying that commit from upstream to see if it solves your problem? If it does we could change this PR to apply it.

Great, it solves =)

PR #359, opened 2020-07-30
win32: Fix IF_S* macros

joaoantoniocardoso:
Currently there are several sources being built for windows that are using `S_IS*` macros that expands to non defined `S_IF*` macros.

Since `S_IFBLK`, `S_IFSOCK` and `S_IFLNK` are not defined by ucrt because they make no sense for windows, we should remove those definitions from evil and change efl sources to not use them on windows.

### DIFF
--- a/src/bin/efreet/efreetd_cache.c
+++ b/src/bin/efreet/efreetd_cache.c
@@ -518,7 +518,7 @@ desktop_changes_monitor_add(const struct stat *st, const char *path)
    const char *monpath = path;
 
    if (eina_hash_find(desktop_change_monitors, path)) return;
-#ifndef _WIN32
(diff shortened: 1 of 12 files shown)


src/bin/efreet/efreetd_cache.c
```diff
@@ -518,7 +518,7 @@ desktop_changes_monitor_add(const struct stat *st, const char *path)
    const char *monpath = path;
 
    if (eina_hash_find(desktop_change_monitors, path)) return;
-#ifndef _WIN32
+#if defined(S_ISLNK) && !defined(_MSC_VER)
```
Coquinho, line 521: This `#ifndef _WIN32` is not ours, so I think it is okay to don't change anything here.

PR #362, opened 2020-07-31
win32: Add popen/pclose

joaoantoniocardoso:


Coquinho:
Do you know if `pclose` could use [eina pipe](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/src/lib/eina/eina_pipe.h) instead? (I'm not sure about it but as I understand files on windows is kinda messy)

joaoantoniocardoso:
> Do you know if `pclose` could use [eina pipe](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/src/lib/eina/eina_pipe.h) instead? (I'm not sure about it but as I understand files on windows is kinda messy)

Hmm I don't understand it very well too but I don't think so, `pclose` is not the same as `close` and windows provides the buildin `_pclose()`/`_popen()` specifically for pipes.

joaoantoniocardoso:
> UUUUh actually I forgot about this warning: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/popen-wpopen?view=vs-2019
> 
> > This API cannot be used in applications that execute in the Windows Runtime. For more information, see [CRT functions not supported in Universal Windows Platform apps](https://docs.microsoft.com/en-us/cpp/cppcx/crt-functions-not-supported-in-universal-windows-platform-apps?view=vs-2019).

Oh, and there is no workaround:
> 
> [Unsupported CRT Functions](https://docs.microsoft.com/en-us/cpp/cppcx/crt-functions-not-supported-in-universal-windows-platform-apps?view=vs-2019#unsupported-crt-functions)
> | _pclose _pipe _popen _wpopen  | Pipe functionality is not available to UWP apps. | No workaround. |

joaoantoniocardoso:
This is not going anywhere, closing.

PR #364, opened 2020-07-31
win32: Avoid compiling examples that use platform-specific functions not available in Windows

joaoantoniocardoso:
Yeah, we are talking specifically about `phtread` here. `usleep` is already available from `Evil`.

Originally 97d5ddbc2f1fe62c0e3da4cc129cb40ddabb3639

### DIFF
--- a/src/examples/elementary/meson.build
+++ b/src/examples/elementary/meson.build
@@ -102,10 +102,6 @@ examples = [
   'web_example_02',
   'win_example',
   'track_example_01',
-  'efl_thread_1',
-  'efl_thread_2',
-  'efl_thread_3',
-  'efl_thread_4',
   'efl_thread_5',
   'efl_thread_6',
   'efl_ui_list_example_1',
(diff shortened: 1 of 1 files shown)


Coquinho:
Should we open an issue/phab ticket about reworking those examples to use eina threads?

joaoantoniocardoso:
> Could we change `src/examples/elementary/meson.build` so we don't build those examples native windows builds?

Indeed. Take a look now.

src/examples/elementary/meson.build
```diff
@@ -123,6 +119,16 @@ examples = [
...
 ]
 
+# Disabling examples that are directly using pthreads 
+if !sys_windows
+  examples += [
+    'efl_thread_1',
+    'efl_thread_2',
+    'efl_thread_3',
+    'efl_thread_4',
+  ]
```
Coquinho, line 129: probably a good idea to use `files`:
```suggestion
  examples += files([
    'efl_thread_1',
    'efl_thread_2',
    'efl_thread_3',
    'efl_thread_4',
  ])
```

Coquinho:
> > Could we change `src/examples/elementary/meson.build` so we don't build those examples native windows builds?
> 
> Indeed. Take a look now.

Yeah, it got much cleaner, thanks.
Oh, linux CI just failed Oo

src/examples/elementary/meson.build
joaoantoniocardoso: They are not files.

Coquinho: ahhh, right right

PR #367, opened 2020-07-31
Add response-file support to eolian_gen

JPTIZ:


src/bin/eolian/main.c
```diff
@@ -516,9 +517,79 @@ init_cmdline_options(CmdLineOptions* options) {
...
+        {
+            case '\n':
+                ++argc;
+            case '\r':
+                contents[i] = '\0';
+            default:
+                break;
+        }
+    }
+    printf("::   Total arguments: %d\n", argc); fflush(stdout);
```
Coquinho, line 567: Is there a need for a flush in every print or is it for debugging?

src/tests/eolian/eolian_generation.c
```diff
@@ -65,18 +65,89 @@ _files_compare (const char *ref_filename, const char *tmp_filename)
...
+    printf("Opening %s...", rsp_path);
+
+    remove(rsp_path);
+    FILE *f = fopen(rsp_path, "w");
+
+    if (!f)
+    {
+        printf("Failed!\n");
+        return -1;
+    }
```
Coquinho, line 83:
```suggestion
    if (!f)
      {
         printf("Failed!\n");
         return -1;
      }
```

src/tests/eolian/eolian_generation.c
```diff
@@ -65,18 +65,89 @@ _files_compare (const char *ref_filename, const char *tmp_filename)
...
+        "%s\n",
+        type, output_filename,
+        eo_filename
+    );
+
+    if (ferror(f))
+        printf("Foi no ferror");
+
+    if (fclose(f))
+        return -1;
```
Coquinho, line 110: You can put those on the same line or:

```suggestion
    if (ferror(f))
      printf("Foi no ferror");

    if (fclose(f))
      return -1;
```

src/tests/eolian/eolian_generation.c
```diff
@@ -65,18 +65,89 @@ _files_compare (const char *ref_filename, const char *tmp_filename)
...
+                 rsp_path);
+
+        int failed = _create_response_file(rsp_path, output_filename, type, eo_filename);
+
+        if (failed)
+        {
+            fprintf(stderr, "Failed to create response file %s\n", rsp_path);
+            perror("Reason");
+            fflush(stderr);
+            return -1;
```
Coquinho, line 142: Are `abort` and `return -1` equivalent semanticaly?

src/bin/eolian/main.c
JPTIZ: Debugging :)

src/tests/eolian/eolian_generation.c
JPTIZ: Just wait </3

src/tests/eolian/eolian_generation.c
JPTIZ: I really am against same-line statements, so I'll stick with the second.

But remind me when I open for review, for now I'm just trying to understand why it crashes here (and why it reads only 1 arg and just ignores the others).

src/tests/eolian/eolian_generation.c
JPTIZ: `abort` is to send a `SIGABORT` to the program. `-1` is just a "things went wrong!" answer to the calling function.

src/tests/eolian/eolian_generation.c
Coquinho: gotta go fast!!!!

src/tests/eolian/eolian_generation.c
Coquinho: okidoki

src/tests/eolian/eolian_generation.c
```diff
@@ -65,17 +65,63 @@ _files_compare (const char *ref_filename, const char *tmp_filename)
...
+   fprintf(f,
+       "-S\n"
+       "-I\n"
+       ""TESTS_SRC_DIR"\n"
+       "-o\n"
+       "%s:%s\n"
+       "%s",
+       type, output_filename,
+       eo_filename
+   );
```
joaoantoniocardoso, line 89: Not an issue but it seems quite odd to me to read, maybe you consider something like the following:
```suggestion
   fprintf(f, "-S\n" "-I\n" TESTS_SRC_DIR "\n" "-o\n" 
             "%s:" "%s\n" "%s",
             type, output_filename, eo_filename);
```

src/bin/eolian/main.c
```diff
@@ -492,47 +493,117 @@ _write_deps(const Eolian_State *eos, const char *ofname, const char *ifname,
...
+   int gen_what;
+   char *basen;
+   char *outs[sizeof(_dexts) / sizeof(void *)];
+   const char *input;
+} Cmd_Line_Options;
+
+void
+init_cmdline_options(Cmd_Line_Options* options)
+{
+   *options = (Cmd_Line_Options){
```
joaoantoniocardoso, line 509:
```suggestion
   *options = (Cmd_Line_Options) {
```

src/bin/eolian/main.c
```diff
@@ -492,47 +493,117 @@ _write_deps(const Eolian_State *eos, const char *ofname, const char *ifname,
...
+      .gen_what = 0,
+      .basen = NULL,
+      .outs = {
+        NULL, NULL, NULL, NULL, NULL, NULL
+      },
+      .input = NULL,
+   };
+}
+
 int
```
joaoantoniocardoso, line 522:
```suggestion
static int
```

src/bin/eolian/main.c
```diff
@@ -492,47 +493,117 @@ _write_deps(const Eolian_State *eos, const char *ofname, const char *ifname,
...
-main(int argc, char **argv)
+process_options(int argc, char** argv, Cmd_Line_Options* options);
+
+void
```
joaoantoniocardoso, line 525:
```suggestion
static void
```

src/bin/eolian/main.c
```diff
@@ -492,47 +493,117 @@ _write_deps(const Eolian_State *eos, const char *ofname, const char *ifname,
...
+          }
+     }
+
+   int old_optind = optind;
+   optind = 0;
+   process_options(argc, argv, options);
+   optind = old_optind;
+}
+
+int
```
joaoantoniocardoso, line 592:
```suggestion
static int
```

src/bin/eolian/main.c
```diff
@@ -492,47 +493,117 @@ _write_deps(const Eolian_State *eos, const char *ofname, const char *ifname,
...
+   Eina_Bool scan_system;
+   Eina_List *includes;
+   int pret;
+   int gen_what;
+   char *basen;
+   char *outs[sizeof(_dexts) / sizeof(void *)];
+   const char *input;
+} Cmd_Line_Options;
+
+void
```
joaoantoniocardoso, line 506:
```suggestion
static void
```

src/tests/eolian/eolian_generation.c
JPTIZ: I don't see a reason for that (although those last two lines do seems strange). This way it is pretty clear that it is a line by line separation.

Also that reminds me of EFL's leading comma.

JPTIZ:
> 
> 
> No big deal with the code, but I think the added functions should be _static_ and maybe prefixed with an underscore.

I indeed didn't realize the `static` there. Good point.

src/tests/eolian/eolian_generation.c
joaoantoniocardoso: Yeah, it is ok, it is more a matter of personal taste rather than something technical. Just saying that to me the way is written is not the clearest way to state what is important on that section of code.

JPTIZ: See if the way I let it now is a little better.

src/bin/eolian/main.c
```diff
@@ -492,47 +493,101 @@ _write_deps(const Eolian_State *eos, const char *ofname, const char *ifname,
...
+   Eina_List *includes;
+   int pret;
+   int gen_what;
+   char *basen;
+   char *outs[sizeof(_dexts) / sizeof(void *)];
+   const char *input;
+} Cmd_Line_Options;
+
+static int
+_process_options(int argc, char** argv, Cmd_Line_Options* options);
```
joaoantoniocardoso, line 507: Two things not so important: (1) declarations should be done in one line and (2) you may prefer `type *var` over `type* var`:
```suggestion
static int _process_options(int argc, char **argv, Cmd_Line_Options *options);
```
Possibly it occur in other places of the code too.

JPTIZ: OH, right, this `type *var` is also another EFL standard I forgot.

JPTIZ: Push'd.

src/bin/eolian/main.c
```diff
@@ -492,47 +493,101 @@ _write_deps(const Eolian_State *eos, const char *ofname, const char *ifname,
...
+   char *basen;
+   char *outs[sizeof(_dexts) / sizeof(void *)];
+   const char *input;
+} Cmd_Line_Options;
+
+static int
+_process_options(int argc, char **argv, Cmd_Line_Options *options);
+
+static void
+_read_options_from_response_file(Cmd_Line_Options* options, const char* filename)
```
joaoantoniocardoso, line 510: Oh, here too:
```suggestion
static int _process_options(int argc, char **argv, Cmd_Line_Options *options);

static void _read_options_from_response_file(Cmd_Line_Options *options, const char *filename)
```

src/bin/eolian/main.c
```diff
@@ -492,47 +493,99 @@ _write_deps(const Eolian_State *eos, const char *ofname, const char *ifname,
...
+   int pret;
+   int gen_what;
+   char *basen;
+   char *outs[sizeof(_dexts) / sizeof(void *)];
+   const char *input;
+} Cmd_Line_Options;
+
+static int _process_options(int argc, char **argv, Cmd_Line_Options *options);
+
+static void _read_options_from_response_file(Cmd_Line_Options *options, const char *filename)
```
joaoantoniocardoso, line 508: :100: 
```suggestion
static void 
_read_options_from_response_file(Cmd_Line_Options *options, const char *filename)
```

JPTIZ: Oh wait NOT WITH THAT TRAILING WHITESPACE

JPTIZ:
```suggestion
static void
_read_options_from_response_file(Cmd_Line_Options *options, const char *filename)
```

src/bin/eolian/main.c
```diff
@@ -492,47 +493,100 @@ _write_deps(const Eolian_State *eos, const char *ofname, const char *ifname,
...
+   int pret;
+   int gen_what;
+   char *basen;
+   char *outs[sizeof(_dexts) / sizeof(void *)];
+   const char *input;
+} Cmd_Line_Options;
+
+static int _process_options(int argc, char **argv, Cmd_Line_Options *options);
+
+static void
```
JPTIZ, line 508:
```suggestion
static void
```

joaoantoniocardoso:
@JPTIZ Travis CI didn't like it very much.

PR #369, opened 2020-07-31
win32: Make emotion optional

Coquinho:
This makes `emotion` optional by making it dependent on `emotion` option.

As now it is optional, this:
- removes is from ignored subprojects;
- set `emotion` option to false for native-windows builds;
- remove it from `elementary` default dependencies and add it when `emotion` option is enabled;
- make related files only compile when `emotion` option is enabled;
- guard related code in `elementary` with `HAVE_EMOTION`;

Note that `HAVE_<LIB>` is already defined at [meson.build](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/meson.build#L563).

Original: 6ddc32b


Coquinho:
Note that `HAVE_EMOTION` is already defined at [meson.build](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/meson.build#L566).

Coquinho:
Conflicts resolved.

Coquinho:
This conflict resolution is kinda trick, so please have a look at this again, just to be sure.

PR #370, opened 2020-08-03
win32: Make ethumb optional

Coquinho:
Original: 3626e00dd1329d56cacab614f380f0facb162a8b

### DIFF
--- a/src/lib/elementary/meson.build
+++ b/src/lib/elementary/meson.build
@@ -849,7 +849,6 @@ elementary_src = files([
   'elm_sys_notify_dbus.c',
   'elm_table.c',
   'elm_theme.c',
-  'elm_thumb.c',
   'elm_toolbar.c',
   'elm_transit.c',
   'elm_util.c',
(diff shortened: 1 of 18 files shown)


Coquinho:
Note that `HAVE_ETHUMB` is already defined at [meson.build](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/meson.build#L566).

src/lib/elementary/meson.build
```diff
@@ -958,10 +957,16 @@ elementary_src = files([
...
 elementary_pub_deps = [eina, eet, evas, ecore, ecore_evas, ecore_file, ecore_input, ecore_imf, ecore_con,
-                       edje, eldbus, efreet, ethumb_client, efl]
+                       edje, eldbus, efreet, efl]
 elementary_ext_deps = [atspi, dl, intl, buildsystem_simple]
+
+if get_option('ethumb')
+  elementary_src += files('elm_thumb.c')
+  elementary_deps += [ethumb, ethumb_client]
+  elementary_pub_deps += ethumb_client
+endif
```
JPTIZ, line 969: I was thinking here, what is more..."mesonic" or would sound better?

1. Initialize elementary_deps without the `ethumb{,_client}` and add it inside the `if get_option('ethumb')` (i.e. how you proposed); or
2. Initialize elementary_deps with `ethumb{,_client}` inside (i.e. how it was before the change), which can be the dependency itself or an empty `declare_dependency`.

Coquinho: hmmm, I prefer the first one, but I'm not much of a meson guy...
If we choose the second we need to change a little `meson.build` to generate the empty declarations. It wouldn't be a big change either, but it could lead us to some hard debugging in the future.

JPTIZ: Hum, fair.

Well, we already have to filter source-files, soooooooooo ok!

Coquinho:
Conflicts resolved.

PR #371, opened 2020-08-03
Make eldbus depended of dbus

Coquinho:
This makes `eldbus` optional by making it dependent on `dbus` option.

As now it is optional, this:
- removes is from ignored subprojects;
- set `eldbus` option to false for native-windows builds;
- remove it and `atspi` from `elementary` default dependencies and add it when `dbus` option is enabled;
- make related files only compile when `dbus` option is enabled;
- guard related code in `elementary` with `HAVE_ELDBUS`;

Note that `HAVE_<LIB>` is already defined at [meson.build](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/meson.build#L563).

Originals:
- 5202192b0569873093d2441e48b6187b6c98d194
- 2e0c6914f6937b33ac423e6a7c1125a53f364dc2

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -100,11 +100,6 @@ add_global_arguments(dev_cflags, language: 'c')
 add_global_arguments(dev_cflags, language: 'cpp')
 
 if sys_windows
-  if not get_option('dbus')
-    dbus = declare_dependency()
-    eldbus = declare_dependency()
-  endif
-
   dl = declare_dependency()
   m = declare_dependency()
 
@@ -439,7 +434,7 @@ subprojects = [
 ['emile'            ,[]                    , false,  true, false, false,  true,  true,  true, ['eina', 'efl'], ['lz4', 'rg_etc']],
 ['eet'              ,[]                    , false,  true,  true, false,  true,  true,  true, ['eina', 'emile', 'efl'], ['rg_etc']],
 ['ecore'            ,[]                    , false,  true, false, false,  true, false,  true, ['eina', 'eo', 'efl'], ['buildsystem']],
-['eldbus'           ,[]                    , false,  true,  true, false,  true,  true,  true, ['eina', 'eo', 'efl'], []],
+['eldbus'           ,['dbus']              , false,  true,  true, false,  true,  true,  true, ['eina', 'eo', 'efl'], []],
 ['ecore'            ,[]                    ,  true, false, false, false, false,  true,  true, ['eina', 'eo', 'efl'], []], #ecores modules depend on eldbus
 ['ecore_audio'      ,['audio']             , false,  true, false, false, false, false,  true, ['eina', 'eo'], []],
 ['ecore_avahi'      ,['avahi']             , false,  true, false, false, false,  true, false, ['eina', 'ecore'], []],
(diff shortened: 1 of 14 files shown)


Coquinho:
Note that `HAVE_ELDBUS` is already defined at [meson.build](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/meson.build#L566).

Coquinho:
Should this PR remove [this if](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/meson.build#L103)?

Coquinho:
> Should this PR remove [this if](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/meson.build#L103)?

Yeah, it should: 2e0c6914f6937b33ac423e6a7c1125a53f364dc2

meson.build
```diff
@@ -472,7 +467,7 @@ subprojects = [
 ['emotion'          ,[]                    ,  true,  true, false, false,  true,  true,  true, ['eina', 'efl', 'eo'], []],
 ['ethumb'           ,['ethumb']            ,  true,  true,  true, false, false, false,  true, ['eina', 'efl', 'eo'], []],
 ['ethumb_client'    ,['ethumb']            , false,  true,  true, false, false,  true,  true, ['eina', 'efl', 'eo', 'ethumb'], []],
-['elementary'       ,[]                    ,  true,  true,  true,  true,  true,  true,  true, ['eina', 'efl', 'eo', 'eet', 'evas', 'ecore', 'ecore-evas', 'ecore-file', 'ecore-input', 'edje', 'emotion', 'ecore-imf', 'ecore-con', 'eldbus', 'efreet', 'efreet-mime', 'efreet-trash', 'eio'], ['atspi']],
+['elementary'       ,[]                    ,  true,  true,  true,  true,  true,  true,  true, ['eina', 'efl', 'eo', 'eet', 'evas', 'ecore', 'ecore-evas', 'ecore-file', 'ecore-input', 'edje', 'emotion', 'ecore-imf', 'ecore-con', 'efreet', 'efreet-mime', 'efreet-trash', 'eio'], ['atspi']],
```
JPTIZ, line 470: To to be clear: you removed from here because it is optional now? Or it doesn't matter here if it is optional or not?

Coquinho, line 470: Should we remove `atspi` from here?

src/bin/elementary/meson.build
```diff
@@ -177,6 +175,13 @@ if get_option('ethumb')
   elementary_test_src += ['test_thumb.c']
 endif
 
+if get_option('dbus')
+  elementary_test_src += [
+    'test_sys_notify.c',
+    'test_systray.c',
+  ]
```
JPTIZ, line 182:
```suggestion
  elementary_test_src += files([
    'test_sys_notify.c',
    'test_systray.c',
  ])
```

src/tests/elementary/meson.build
```diff
@@ -110,6 +109,10 @@ if get_option('ethumb')
   elementary_suite_src += ['elm_test_thumb.c']
 endif
 
+if get_option('dbus')
+  elementary_suite_src += ['elm_test_atspi.c']
```
JPTIZ, line 113:
```suggestion
  elementary_suite_src += files('elm_test_atspi.c')
```

meson.build
Coquinho: Removed because it is optional, it is added back at `elementary ``meson.build` when `dbus` is enabled

src/tests/elementary/meson.build
```diff
@@ -174,6 +176,10 @@ efl_ui_suite_src = [
   'custom_recognizer2.c',
 ]
 
+if get_option('dbus')
+  efl_ui_suite_src += ['efl_ui_test_atspi.c']
```
JPTIZ, line 180:
```suggestion
  efl_ui_suite_src += files('efl_ui_test_atspi.c')
```

meson.build
JPTIZ: Ok.

src/bin/elementary/meson.build
Coquinho: `elementary_test_src` don't use files, it is only a list of names

src/tests/elementary/meson.build
Coquinho: `elementary_suite_src` don't use files, it is only a list of names

src/tests/elementary/meson.build
Coquinho: `efl_ui_suite_src` don't use files, it is only a list of names

src/bin/elementary/meson.build
JPTIZ: ? But it is used right there as a list of sources, line 181.

JPTIZ: Actually there's a plenty of missing `files(...)` in this file. _Oh lord.png_

Coquinho: exactly, I didn't use files, because the variable isn't a `depend_files` and I don't know if it would be best to change it in this PR. Seems like a good thing for a phab ticket.

JPTIZ: `files` is not intended for a specific use case, it is intended to be used with file references. This _prevents_ bugs, so the one we had before was more of a consequence of not using it than a use case. Well, since I'm already messing with meson.build files, I could provide a patch for that later.

JPTIZ: There you go: https://phab.enlightenment.org/T8784.

Coquinho: Nice just added something there.

meson.build
joaoantoniocardoso: I think so, at least will be more coherent with `src/lib/elementary/meson.build `.

Coquinho: hummmm, I've just had a better look at [meson.build](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/meson.build#L584):
```
    foreach static_lib : package[10]
      if get_variable(static_lib, tmp_empty) == tmp_empty
        subdir(join_paths('src', 'static_libs', static_lib))
      endif
    endforeach
```
I don't think we should remove it there, it would clash with how our build system is set now

PR #372, opened 2020-08-03
win32: Do not fail if we can't find libcurl

Coquinho:
Original: c2c869789d038c87657974395b497c3bea67b971

### DIFF
--- a/src/lib/ecore_con/ecore_con_url_curl.c
+++ b/src/lib/ecore_con/ecore_con_url_curl.c
@@ -275,11 +275,7 @@ _c_init(void)
    LOAD("curllib.dll"); // if fail try 2
    if (!_c->mod)
      {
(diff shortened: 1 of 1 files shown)


JPTIZ:
I'm not sure that's the way to go. If we do not have libcurl, what should we be doing?

(You can starting by saying where it is used/what for.)

JPTIZ:
> 
> 
> I'm not sure that's the way to go. If we do not have libcurl, what should we be doing?
> 
> (You can starting by saying where it is used/what for.)

As explained in our meeting: curl is used when paths are specified to URL strings, so it is very contextual and not necessary at start time. What makes more sense is:
- Give the "missing libcurl" warning during EFL startup instead of erroring (which is what is happening in this PR);
- Give an error message (but do not crash) when trying to use an URL path in runtime.

PR #373, opened 2020-08-04
[wrap] Add and enable libpng

joaoantoniocardoso:
Originally:
 - c74e968991c663f23f67118050630b8c4e684670
 - b09db64585cf5234227470caf1a9dd1789d2df11
 - 5bafbf139134a5952a6780dfb6a48dce14d8dce8
 - 6001e0a3534af31a0387459170146daf96cb631c


src/lib/evas/meson.build
```diff
@@ -4,11 +4,9 @@
...
 #
 
-if not sys_windows
-  png = dependency('libpng')
-else
-  png = declare_dependency()
-endif
+png = dependency('libpng',
+  fallback: ['libpng', 'png_dep']
+)
```
Coquinho, line 9: I think there is no need to use 3 lines here:
```suggestion
png = dependency('libpng', fallback: ['libpng', 'png_dep'])
```

.gitignore
```diff
@@ -101,4 +101,6 @@ env.bat
 
 # Subprojects
 subprojects/*
+!subprojects/packagefiles
+!subprojects/packagefiles/*
```
Coquinho, line 105: Why ignoring package files if this PR addsit?

joaoantoniocardoso: My bad, that shouldn't be committed.

Coquinho: ohhhhh, right right, I didn't note the `!`, thanks.

PR #375, opened 2020-08-04
elm: Use platform neutral embed_config C program instead of script

joaoantoniocardoso:
This PR fixes the following error when trying to build elementary:
```
FAILED: src/lib/elementary/elementary-1.dll.p/elm_config.c.obj
<...>
In file included from ../src/lib/elementary/elm_config.c:1706:
src/lib/elementary\elm_default_config.x(1,1): error: expected identifier or '('
"static const char *embedded_config = \"\""
^
```

### DIFF
--- a/src/bin/elementary/embed_config.c
+++ b/src/bin/elementary/embed_config.c
@@ -1,4 +1,3 @@
-
 #include <assert.h>
 #include <errno.h>
 #include <getopt.h>
@@ -7,27 +6,27 @@
 #include <string.h>
 
 #ifdef HAVE_CONFIG1
-#include "elementary_config.h"
+# include "elementary_config.h"
 #endif
 
 static void
-_print_usage(const char *progn, FILE *outf)
+_print_usage(const char *progn)
 {
-   fprintf(outf, "Usage: %s [options] [input]\n", progn);
-   fprintf(outf, "Options:\n"
+   fprintf(stdout, "Usage: %s [options] [input]\n", progn);
+   fprintf(stdout, "Options:\n"
                  "  -o output     specify the filename and path for output\n"
                  "  -v            show version number\n"
                  "\n");
 }
 
-char* _read_input_file (FILE* f, const char* fname)
+char *
+_read_input_file(FILE* f, const char* fname)
 {
    fseek(f, 0, SEEK_END);
    long fs = ftell(f);
    if (fs < 0)
      {
         fprintf(stderr, "eolian: could not get length of '%s'\n", fname);
-        fclose(f);
         return NULL;
      }
    fseek(f, 0, SEEK_SET);
@@ -36,7 +35,7 @@ char* _read_input_file (FILE* f, const char* fname)
    if (!cont)
      {
         fprintf(stderr, "eolian: could not allocate memory for '%s'\n", fname);
-        fclose(f);
+        free(cont);
         return NULL;
      }
 
@@ -46,16 +45,15 @@ char* _read_input_file (FILE* f, const char* fname)
         fprintf(stderr, "eolian: could not read %ld bytes from '%s' (got %ld)\n",
                 fs, fname, as);
         free(cont);
-        fclose(f);
         return NULL;
      }
 
    cont[fs] = '\0';
-   fclose(f);
    return cont;
 }
 
-int main(int argc, char** argv)
+int
+main(int argc, char** argv)
 {
    char *output = NULL, *input = NULL;
    FILE *fin = NULL, *fout = NULL;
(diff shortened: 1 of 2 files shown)


Coquinho:
Travis didn't liked it

src/bin/elementary/embed_config.c
```diff
@@ -0,0 +1,149 @@
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <getopt.h>
+#include <string.h>
+#include <assert.h>
```
JPTIZ, line 6: 1. I would reorder alphabetically.
2. This:

```suggestion
#include <assert.h>
#include <errno.h>
#include <getopt.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
```

So....

```suggestion
#include <stdio.h>
#include <stdlib.h>
#include <getopt.h>
#include <string.h>
#include <assert.h>
```

Coquinho:
@joaoantoniocardoso Travis broke :/

src/bin/elementary/embed_config.c
```diff
@@ -0,0 +1,150 @@
...
+   for (int opt; (opt = getopt(argc, argv, "ho:")) != -1;)
+     switch (opt)
+       {
+        case 0:
+          break;
+        case 'o':
+          output = strdup(optarg);
+          break;
+        case 'h':
+          _print_usage(argv[0], stdout);
```
Coquinho, line 73: if this will always print on stdout, do we need to set an arg to which file stream to output to?

src/bin/elementary/embed_config.c
```diff
@@ -36,7 +35,8 @@ char* _read_input_file (FILE* f, const char* fname)
    if (!cont)
      {
         fprintf(stderr, "eolian: could not allocate memory for '%s'\n", fname);
-        fclose(f);
+        free(cont);
+        cont = NULL;
```
Coquinho, line 39: Why set `count` to `NULL?

src/lib/elementary/meson.build
```diff
@@ -262,16 +262,15 @@ endforeach
...
-  embed_script = find_program('config_embed.bat')
-else
-  embed_script = find_program('config_embed')
-endif
+embed_config_exe = executable('embed_config_exe', files('../../bin/elementary/embed_config.c'), dependencies: getopt)
+meson.override_find_program('embed_config', embed_config_exe)
+
+embed_script_run = find_program('embed_config', required: false)
 
-embed_config = custom_target('create_embedded_default_config',
```
Coquinho, line 271: Is there any reason to change this name?

joaoantoniocardoso: only thing came to my mind is that it is not a _script_ but an executable? I dont know

Coquinho: welp...I think at least the `embed_config` should stay with the same name, but don't have an opinion about the script

JPTIZ: Actually this is not the executable, this is the "inline-header" generator, for a "config.h-like" header. See:

```meson
  output: 'elm_default_config.x',
  command: [embed_script, '@INPUT@', '@OUTPUT@']
```

So it uses the `embed_script_run` (which is the executable `embed_config_exe`) to generate the header (just like those `configuration_data` + `configure` to generate config.h, but using our executable to generate it instead of a sed-like internal procedure from meson). See the `.in` file used as input?

But this seems a bit overcomplicated. I don't see a reason (looking quickly) to override the `find_program` for `embed_config` (that `meson.override_find_program(a, b)` means "when you use `find_program(program name)` with `a` as program name, use the target `b` instead of anything it could find). Maybe `custom_target` doesn't accept meson targets in `command`?

JPTIZ: Ok, it actually can (see the `command` kwarg in [Meson docs about `custom_target`s](https://mesonbuild.com/Reference-manual.html#custom_target)), so this whole code could've be...

JPTIZ:
```meson
embed_config = executable('embed_config_exe',
    files('../../bin/elementary/embed_config.c'),
    dependencies: getopt)

elementary_default_config = custom_target('create_embedded_default_config',
  input: join_paths(meson.source_root(), 'data', 'elementary', 'config', 'standard', 'base.src.in'),
  output: 'elm_default_config.x',
  command: [embed_script_run, '-o', '@OUTPUT@', '@INPUT@'],
)

**jumps to line 1026**

elementary_lib = library('elementary',
    elementary_src, pub_eo_file_target, priv_eo_file_target, c, elementary_default_config,
    dependencies: elementary_pub_deps + elementary_deps + elementary_ext_deps,
    include_directories : config_dir + [include_directories('.')] + [include_directories(join_paths('..', '..', '..'))],
    install: true,
    c_args : [elm_package_c_args, '-DELM_BUILD'],
    version : meson.project_version()
)
```

JPTIZ: To be honest, IMO `embed_config` should be generated in `src/bin/elementary`, since there's where the source actually is, and not here.

src/bin/elementary/embed_config.c
```diff
@@ -0,0 +1,147 @@
...
+
+#ifdef HAVE_CONFIG1
+# include "elementary_config.h"
+#endif
+
+static void
+_print_usage(const char *progn)
+{
+   fprintf(stdout, "Usage: %s [options] [input]\n", progn);
+   fprintf(stdout, "Options:\n"
```
JPTIZ, line 16:
```suggestion
   printf("Usage: %s [options] [input]\n", progn);
   printf("Options:\n"
```

src/lib/elementary/meson.build
joaoantoniocardoso: > To be honest, IMO `embed_config` should be generated in `src/bin/elementary`, since there's where the source actually is, and not here.

The generator is from `src/bin/elementary` but the sources it is using are from `src/lib/elementary`, the place seems to be right to me.

JPTIZ: Oh, ok then.

PR #376, opened 2020-08-04
win32: Add -in-tree option in edje_cc for Windows

joaoantoniocardoso:
This PR fixes the following error when building elementary:
```
[2871/3123] Generating edje_cc_lua_script.edc with a meson_exe.py custom command
FAILED: src/examples/edje/lua_script.edj
edje_cc.exe: Error. Cannot run epp: c:/lib/edje/utils/v-1.24/epp.exe
```


src/tests/edje/tests/meson.build
```diff
@@ -6,8 +6,8 @@ foreach edc_file : edc_files
    themes += custom_target('edje_cc_' + edc_file,
     input : edc_file,
     output : '@BASENAME@.edj',
-    command : edje_cc_exe + [ '-beta', '-fastcomp',
-              '-id', join_paths(meson.source_root(), 'src', 'tests', 'emotion', 'data'),
+    command : edje_cc_exe + [ '-beta', '-fastcomp', '-in-tree',
+              '-id', join_paths(meson.source_root(), 'src', 'tests', 'data'),
```
JPTIZ, line 10: ?

src/bin/edje/edje_cc.c
```diff
@@ -195,6 +197,11 @@ main(int argc, char **argv)
           {
              eina_log_domain_level_set("edje_cc", EINA_LOG_LEVEL_INFO);
           }
+        else if (!strcmp(argv[i], "-in-tree"))
+          {
+             putenv("EFL_RUN_IN_TREE=1");
+             in_tree = 1;
```
JPTIZ, line 203: Maybe we could try a slightly different approach: we rely on `in_tree` global variable (which could be renamed to `efl_run_in_tree`) and, in the start of `main`, check for `getenv(EFL_RUN_IN_TREE)` is not 0 and if that's the case then set `efl_run_in_tree` to 1.

src/tests/edje/tests/meson.build
JPTIZ: Ok, just a plain "?" isn't a proper way to ask things, so...

Why exactly are we removing `emotion` there?

src/bin/edje/edje_cc.c
Coquinho: Sounds better for me too.

src/tests/edje/tests/meson.build
joaoantoniocardoso: Hmm yeah, that seems wrong to me. I will take a look at all the other places too

src/bin/edje/edje_cc.c
joaoantoniocardoso: Nice, I'll try that.

src/tests/edje/tests/meson.build
Coquinho: Maybe this should be at #369 ?

src/tests/edje/data/meson.build
```diff
@@ -25,7 +25,8 @@ foreach edc_file : edc_files
    themes += custom_target('edje_cc_' + edc_file,
     input : edc_file,
     output : '@BASENAME@.edj',
-    command : edje_cc_exe + [ '-beta', '-fastcomp',
+    command : edje_cc_exe + [ '-beta', '-fastcomp', '-in-tree',
+              '-id', join_paths(meson.source_root(), 'src', 'tests', 'data'),
```
joaoantoniocardoso, line 29: @JPTIZ here it is another odd change

Coquinho: this probably should be to remove emotion too

src/tests/edje/tests/meson.build
joaoantoniocardoso: actually there is no `src/tests/data` in tree.

src/tests/edje/data/meson.build
joaoantoniocardoso: > actually there is no src/tests/data in tree, I'll keep only src/tests/emotion/data here and everywhere else.

PR #378, opened 2020-08-07
eet: Fix 'No OPENSSL_Applink' error message.

JPTIZ:
In some systems, OpenSSL raises an error about "No Applink" (see ["I've compiled a program under Windows and it crashes: why?" in OpenSSL FAQ](https://www.openssl.org/docs/faq.html#PROG3)). Including only openssl/applink.c didn't work, so the solution was to replace `FILE*` interfaces with OpenSSL's BIO API which also contains file operations.

Notice that this doesn't solve Eet fully, since now that applink works (for this case), there are tests failing.

### DIFF
--- a/src/lib/eet/eet_cipher.c
+++ b/src/lib/eet/eet_cipher.c
@@ -190,21 +190,19 @@ on_error:
 
 # else /* ifdef HAVE_GNUTLS */
    /* Openssl private declarations */
-   FILE *fp;
    EVP_PKEY *pkey = NULL;
    X509 *cert = NULL;
 
    if (!emile_cipher_init()) return NULL;
 
    /* Load the X509 certificate in memory. */
-   fp = fopen(certificate_file, "rb");
-   if (!fp)
-     return NULL;
-
-   cert = PEM_read_X509(fp, NULL, NULL, NULL);
-   fclose(fp);
-   if (!cert)
-     goto on_error;
+   {
+      BIO* cert_bio = BIO_new_file(certificate_file, "rb");
+      cert = PEM_read_bio_X509(cert_bio, NULL, NULL, NULL);
+      BIO_free(cert_bio);
+      if (!cert)
+        goto on_error;
+   }
 
    /* Check the presence of the public key. Just in case. */
    pkey = X509_get_pubkey(cert);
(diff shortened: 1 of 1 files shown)


src/lib/eet/eet_cipher.c
```diff
@@ -190,36 +190,33 @@ eet_identity_open(const char               *certificate_file,
...
-   fclose(fp);
-   if (!cert)
-     goto on_error;
+   {
+      BIO* cert_bio = BIO_new_file(certificate_file, "rb");
+      cert = PEM_read_bio_X509(cert_bio, NULL, NULL, NULL);
+      BIO_free(cert_bio);
+      if (!cert)
+        goto on_error;
+   }
```
Coquinho, line 205: do you know if this will work too for cygwin?

Coquinho: maybe we could guard to make this way only for `_MSC_VER`?

src/lib/eet/eet_cipher.c
```diff
@@ -190,36 +190,33 @@ eet_identity_open(const char               *certificate_file,
...
-   fclose(fp);
-   if (!pkey)
-     goto on_error;
+   {
+      BIO* private_key_bio = BIO_new_file(private_key_file, "rb");
+      pkey = PEM_read_bio_PrivateKey(private_key_bio, NULL, cb, NULL);
+      BIO_free(private_key_bio);
+      if (!pkey)
+        goto on_error;
+   }
```
Coquinho, line 219: Why surrounding it with `{}`?

src/lib/eet/eet_cipher.c
JPTIZ: > do you know if this will work too for cygwin?

Of course it will, this is OpenSSL's API.

> maybe we could guard to make this way only for `_MSC_VER`?

There's no need for it, this isn't a platform-dependent code in any form.

src/lib/eet/eet_cipher.c
JPTIZ: To ensure variables declared here stay here. More of a hint of "this piece of code have temporary values on it that shouldn't be used later".

PR #380, opened 2020-08-07
Evas: Call evas modules by its final name in native-windows builds

Coquinho:
Different from the original intention, this doesn't make modules names
work the same in Linux and windows. This workaround the need for a
script for native windows builds but lets other platforms work as they
already do. This only change module names for its final version
(`module`) when on native windows builds. To achieve this, it creates a
`sys_native_windows` variable, which is only true when using `msvc` or
`clang-cl` while in a `windows` host.

Original: 2f3c8295db3e1fa9c62462dbeb6057085e19f87e

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -50,9 +50,10 @@ osx = ['darwin']
 sys_linux = linux.contains(host_machine.system())
 sys_bsd = bsd.contains(host_machine.system())
 sys_windows = windows.contains(host_machine.system())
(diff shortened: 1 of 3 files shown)


Coquinho:
This seems to don't impact at compilarion, so I'm putting this assid for now.

Coquinho:
Okay, reviving this PR: https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/src/lib/ecore_evas/ecore_evas_module.c#L26

PR #384, opened 2020-08-13
Eet: Resolve eet_test_cache test case

Coquinho:
This test case had 2 problems one while creating a temporary file and other
while closing it:

About creating the file: `/tmp` doesn't exist in windows and `mkstemp` expects
that all path given already exists (besides the file it will create), causing a
fail:
```
74:    fail_if(-1 == (tmpfd = mkstemp(file)));
```
Here this is solved by using `eina_file_mkstemp` that creates a temporary file
at a temporary directory obtained with `eina_envitonment_tmp` which is an
architecture-independent function.

About closing it: `eet_close` is a postponed close, it doesn't immediately close
files opened as `write and read` nor `write` unless it has non saved changes nor
immediately closes files opened as `read` files. So when arriving at `unlink`
the file may not be closed yet. 
On linux it isn't a problem, as when `unlink`ing an open file it returns `0` and
waits until all references to the file to be closed for its removal.
But on windows while `unlink`ing an open file `-1` is returnged,  causing a fail
at:
```
102:  fail_if(unlink(file) != 0);
```
With `unlink` setting `errno` to `13`(`strerror(13)` = `Permission denied`).

This is solved using `eet_clearcache` which forces the closure of every file (in
case the file has been altered it flushes it before closing).
Note that this solves the problem for *this* test, but it isn't the best
solution, an user still needs to know where his program will run to know if they
need to force file closure before  `unlink`.

### DIFF
--- a/src/tests/eet/eet_test_cache.c
+++ b/src/tests/eet/eet_test_cache.c
@@ -60,7 +60,10 @@ EFL_START_TEST(eet_test_cache_concurrency)
    Eina_Bool r;
    int tmpfd;
 
-   file = strdup("/tmp/eet_suite_testXXXXXX");
+   int path_size = strlen(eina_environment_tmp_get()) + strlen("eet_suite_testXXXXXX") + 2;
+   file = malloc(sizeof(char)*path_size);
+   eina_file_path_join(file, path_size
+      , eina_environment_tmp_get(), "eet_suite_testXXXXXX");
 
    eina_threads_init();
 
@@ -94,7 +97,7 @@ EFL_START_TEST(eet_test_cache_concurrency)
    thread_ret = eina_thread_join(thread);
    fail_unless(thread_ret == NULL, (char const *)thread_ret);
 
-   eet_close(ef);
+   eet_clearcache();
 
    fail_if(unlink(file) != 0);
 
--- a/src/tests/eet/eet_test_cache.c
+++ b/src/tests/eet/eet_test_cache.c
@@ -5,7 +5,7 @@
 #include <stdlib.h>
 
 #ifndef _WIN32
-#include <unistd.h>
+# include <unistd.h>
 #else
 # include <evil_private.h>
 #endif
@@ -51,22 +51,14 @@ _open_close_worker(void *path, Eina_Thread tid EINA_UNUSED)
 
 EFL_START_TEST(eet_test_cache_concurrency)
 {
-   char *file;
    const char *buffer = "test data";
    Eet_File *ef;
    void *thread_ret;
    unsigned int n;
    Eina_Thread thread;
+   Eina_Tmpstr *tmpfile = NULL;
    Eina_Bool r;
    int tmpfd;
-
-   const char * filename = "eet_suite_testXXXXXX";
-   const char * tmpdir = eina_environment_tmp_get();
-   // +2 stands for <path separator> + <end of string>
-   size_t path_size = strlen(tmpdir) + strlen(filename) + 2;
-
-   file = malloc(sizeof(char)*path_size);
-   eina_file_path_join(file, path_size , tmpdir, filename);
 
    eina_threads_init();
 
@@ -74,16 +66,18 @@ EFL_START_TEST(eet_test_cache_concurrency)
    eina_condition_new(&open_worker_cond, &open_worker_mutex);
 
    /* create a file to test with */
-   fail_if(-1 == (tmpfd = mkstemp(file)));
+   /* tmpfile will be created in temporary directory (with eina_environment_tmp) */
+   tmpfd = eina_file_mkstemp("eet_suite_testXXXXXX", &tmpfile);
+   fail_if(-1 == tmpfd);
    fail_if(!!close(tmpfd));
-   ef = eet_open(file, EET_FILE_MODE_WRITE);
+   ef = eet_open(tmpfile, EET_FILE_MODE_WRITE);
    fail_if(!ef);
    fail_if(!eet_write(ef, "keys/tests", buffer, strlen(buffer) + 1, 0));
 
    eina_lock_take(&open_worker_mutex);
    /* start a thread that repeatedly opens and closes a file */
    open_worker_stop = 0;
-   r = eina_thread_create(&thread, EINA_THREAD_NORMAL, -1, _open_close_worker, file);
+   r = eina_thread_create(&thread, EINA_THREAD_NORMAL, -1, _open_close_worker, tmpfile);
    fail_unless(r);
 
    eina_condition_wait(&open_worker_cond);
@@ -102,9 +96,10 @@ EFL_START_TEST(eet_test_cache_concurrency)
 
    eet_clearcache();
 
-   fail_if(unlink(file) != 0);
+   fail_if(unlink(tmpfile) != 0);
 
    eina_threads_shutdown();
+   eina_tmpstr_del(tmpfile);
 }
 EFL_END_TEST
 
--- a/src/tests/eet/eet_test_cache.c
+++ b/src/tests/eet/eet_test_cache.c
@@ -95,6 +95,10 @@ EFL_START_TEST(eet_test_cache_concurrency)
    fail_unless(thread_ret == NULL, (char const *)thread_ret);
 
    eet_close(ef);
+    /* As `eet_close` is a postponed close and windows' `unlink` doesn't execute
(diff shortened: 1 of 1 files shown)


src/tests/eet/eet_test_cache.c
```diff
@@ -60,7 +60,10 @@ EFL_START_TEST(eet_test_cache_concurrency)
    Eina_Bool r;
    int tmpfd;
 
-   file = strdup("/tmp/eet_suite_testXXXXXX");
+   int path_size = strlen(eina_environment_tmp_get()) + strlen("eet_suite_testXXXXXX") + 2;
```
JPTIZ, line 63: Why +2?

Coquinho: it uses `\\` to join paths

Coquinho: oh, but it won't use `\\` on every system....is there an workarround?

Coquinho:
[here](https://github.com/expertisesolutions/efl/tree/devs/coquinho/eet/test_cache-show-test) should be easier to see the tests that work actually passing.

src/tests/eet/eet_test_cache.c
joaoantoniocardoso: Did you mean something like `EINA_PATH_SEP`?

JPTIZ: > 
> 
> > Why +2?
> 
> path separator char + end of string char?
> 
> > oh, but it won't use `\\` on every system....is there an workarround?
> 
> Did you mean something like `EINA_PATH_SEP_C`?

`EINA_PATH_SEP_C` seems fair.

Also, htinking now: `\\` is not 2 characters, but 1. So the length should be increased by one. What if we saw what was the string before and after the join?

Coquinho: > Did you mean something like EINA_PATH_SEP_C?

yeah, will try with `+ strlen(EINA_PATH_SEP_C)`, thanks @joaoantoniocardoso.

> Also, htinking now: \\ is not 2 characters, but 1. So the length should be increased by one. What if we saw what was the string before and after the join?

I will look at it, but if i remember it right I've tried it at the begining. Anyway, I will try.

JPTIZ: Thinking again: `EINA_PATH_SEP_C` is a `char` (while `EINA_PATH_SEP` is a `char*`/string, but it has the length of 1 character anyway), so the length would still be 1.

Coquinho: oh sure, sure, `EINA_PATH_SEP_C` is a `char` and `EINA_PATH_SEP_S` is a `string` right, right.

About the sizes, I've run this on linux and windows:
```
   int path_size = strlen(eina_environment_tmp_get()) + strlen("eet_suite_testXXXXXX") + 2;
   file = malloc(sizeof(char)*path_size);
   eina_file_path_join(file, path_size
      , eina_environment_tmp_get(), "eet_suite_testXXXXXX");
fprintf(stderr, "+2 strlen(%s) = %llu\n", file, strlen(file));fflush(stderr);

free(file);
path_size = strlen(eina_environment_tmp_get()) + strlen("eet_suite_testXXXXXX") + 1;
file = malloc(sizeof(char)*path_size);
eina_file_path_join(file, path_size
  , eina_environment_tmp_get(), "eet_suite_testXXXXXX");
fprintf(stderr, "+1 strlen(%s) = %llu\n", file, strlen(file));fflush(stderr);
```
And in both cases we need the +2 to print all the path, we can see that with `+1` one of the `X` is cropped out:
```
+2 strlen(C:\Users\coku\AppData\Local\Temp\eet_suite_testXXXXXX) = 53
+1 strlen(C:\Users\coku\AppData\Local\Temp\eet_suite_testXXXXX) = 52
```
I'm pretty sure that @joaoantoniocardoso was right:
>path separator char + end of string char?

JPTIZ: Hmm just to confirm: what was the value of `eina_environment_tmp_get()`?

JPTIZ: Also, I don't believe it is an end of string, see:

> ```
> +2 strlen(C:\Users\coku\AppData\Local\Temp\eet_suite_testXXXXXX) = 53
> +1 strlen(C:\Users\coku\AppData\Local\Temp\eet_suite_testXXXXX) = 52
> ```

If it was a matter of end-of-string char, it would actually happend that +1 would print _more_ stuff, not less, since it wouldn't copy the end of string character. Since it has shown less, that means the last `X` is the char at `strlen-1` index, and not a `\0`.

JPTIZ: _\*sigh*_ forget what I said: I was thinking about the _displayed_ string length, but not about what the join would do internally.

Give a thumbs up to @Coquinho.

joaoantoniocardoso: hmm, path_size need to accommodate eina_environment_tmp_get() + EINA_PATH_SEP_S + "eet_suite_testXXXXXX" + "\0", so what is wrong with `strlen(eina_environment_tmp_get()) + strlen(EINA_PATH_SEP_S) + strlen("eet_suite_testXXXXXX") + 1`, or just `strlen(eina_environment_tmp_get()) + strlen("eet_suite_testXXXXXX") + 2`?

JPTIZ: Nothing, I was understanding the wrong thing here.

src/tests/eet/eet_test_cache.c
```diff
@@ -60,7 +60,10 @@ EFL_START_TEST(eet_test_cache_concurrency)
...
+   file = malloc(sizeof(char)*path_size);
+   eina_file_path_join(file, path_size
+      , eina_environment_tmp_get(), "eet_suite_testXXXXXX");
```
joaoantoniocardoso, line 66: Not a problem but I would suggest it to be more clear:
```suggestion
   const char *filename = "eet_suite_testXXXXXX";
   char *dirname = eina_environment_tmp_get();
   int file_length = strlen(dirname) + strlen(filename) + 2; // <path separator> + <end of string>
	
   file = malloc(sizeof(char) * file_length);
   eina_file_path_join(file, file_length, dirname, filename);
```

JPTIZ: Not only clear, but less hardcoded strings (which mean more consistency in maintance).

Coquinho: Done, please have a look and let me know if there is anything wrong.

src/tests/eet/eet_test_cache.c
```diff
@@ -60,7 +60,13 @@ EFL_START_TEST(eet_test_cache_concurrency)
    Eina_Bool r;
    int tmpfd;
 
-   file = strdup("/tmp/eet_suite_testXXXXXX");
+   const char * filename = "eet_suite_testXXXXXX";
+   const char * tmpdir = eina_environment_tmp_get();
```
joaoantoniocardoso, line 64: Most of efl seems to prefer `type *var` over `type* var` or `type * var`. The last one looks like a multiplication operator =x

src/tests/eet/eet_test_cache.c
```diff
@@ -94,11 +94,12 @@ EFL_START_TEST(eet_test_cache_concurrency)
    thread_ret = eina_thread_join(thread);
    fail_unless(thread_ret == NULL, (char const *)thread_ret);
 
-   eet_close(ef);
+   eet_clearcache();
```
joaoantoniocardoso, line 97:
```suggestion
   eet_clearcache();
   eet_close(ef);
```
We still need to close that file, otherwise:
```
Running suite(s): Eet
ERR:eet ../src/lib/eet/eet_lib.c:636 eet_shutdown() File 'C:\Users\coku\AppData\Local\Temp\eet_suite_testzWBdba' is still open 1 times !
```

JPTIZ: And this way it passes the test?

joaoantoniocardoso: > And this way it passes the test?

In both ways it passes, but without closing the file it throws that error.

Coquinho: sure, because it needs an `eet_close` to every `eet_open` so `eet` sees it as even.

Coquinho: Done.

PR #385, opened 2020-08-17
Eet: Resolve eet_test_image test case

Coquinho:
Like others `eet` test cases, this test case had 2 problems one while creating a
temporary file and other while closing it:

About creating the file: `/tmp` doesn't exist in windows and `mkstemp` expects
that all path given already exists (besides the file it will create), causing a
fail:
```
../src/tests/eet/eet_test_image.c:93:F:Eet Image:eet_test_image_normal:0: Failure '-1 == (tmpfd = mkstemp(file))' occurred
../src/tests/eet/eet_test_image.c:459:F:Eet Image:eet_test_image_small:0: Failure '-1 == (tmpfd = mkstemp(file))' occurred
```
Here this is solved by using `eina_file_mkstemp` that creates a temporary file
at a temporary directory obtained with `eina_envitonment_tmp` which is an
architecture-independent function.

About closing it: `eet_close` is a postponed close, it doesn't immediately close
files opened as `write and read` nor `write` unless it has non saved changes nor
immediately closes files opened as `read` files. So when arriving at `unlink`
the file may not be closed yet. 
On linux it isn't a problem, as when `unlink`ing an open file it returns `0` and
waits until all references to the file to be closed for its removal.
But on windows while `unlink`ing an open file `-1` is returnged,  causing a fail
at:
```
../src/tests/eet/eet_test_image.c:438:F:Eet Image:eet_test_image_normal:0: Failure 'unlink(file) != 0' occurred
../src/tests/eet/eet_test_image.c:497:F:Eet Image:eet_test_image_small:0: Failure 'unlink(file) != 0' occurred
```
With `unlink` setting `errno` to `13`(`strerror(13)` = `Permission denied`).

This is solved using `eet_clearcache` which forces the closure of every file (in
case the file has been altered it flushes it before closing).
Note that this solves the problem for *this* test, but it isn't the best
solution, an user still needs to know where his program will run to know if they
need to force file closure before  `unlink`.

### DIFF
--- a/src/tests/eet/eet_test_image.c
+++ b/src/tests/eet/eet_test_image.c
@@ -428,6 +428,10 @@ EFL_START_TEST(eet_test_image_normal)
    free(data);
 
    eet_close(ef);
+   /* As `eet_close` is a postponed close and windows' `unlink` doesn't execute
+    * successfully if there is any reference to the file, here `eet_clearcache` is
+    * used to assure that the file is really closed when the unlink happens.
+    */
    eet_clearcache();
 
    fail_if(unlink(tmpfile) != 0);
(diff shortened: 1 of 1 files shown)


Coquinho:
[devs/coquinho/eet_test_image-show](https://github.com/expertisesolutions/efl/tree/devs/coquinho/eet_test_image-show) has a better show case.

Coquinho:
Note that I'm not sure about this approach (about the second problem) because this isn't the cache unity test.

But I've decided to bring it for discussion.

PR #386, opened 2020-08-17
win32: Add --in-tree parameter for edje_codegen and elementary_codegen

Coquinho:
This is needed because Windows platform doesn't have a good way to pass
environment variables to executables through meson, so we pass through a
command-line instead.

As `ecore_evas` and `edje` needs in-tree here they are initialize after
in-tree parameter check.

Original:
- f5b3fd6f3a
- a0208c36b6

### DIFF
--- a/src/bin/edje/edje_codegen.c
+++ b/src/bin/edje/edje_codegen.c
@@ -363,6 +363,7 @@ const Ecore_Getopt optdesc = {
    {
       ECORE_GETOPT_STORE_STR('p', "prefix", "The prefix for the " \
                                             "generataed code."),
+      ECORE_GETOPT_STORE_BOOL('t', "in-tree", "Same as setting EFL_RUN_IN_TREE=1 as environment variable"),
       ECORE_GETOPT_LICENSE('L', "license"),
       ECORE_GETOPT_COPYRIGHT('C', "copyright"),
       ECORE_GETOPT_VERSION('V', "version"),
@@ -1093,10 +1094,12 @@ int
 main(int argc, char *argv[])
 {
    Eina_Bool quit_option = EINA_FALSE;
+   Eina_Bool in_tree_option = EINA_FALSE;
    char *source, *header;
    int arg_index, ret = 0;
    Ecore_Getopt_Value values[] = {
       ECORE_GETOPT_VALUE_STR(prefix),
+      ECORE_GETOPT_VALUE_BOOL(in_tree_option),
       ECORE_GETOPT_VALUE_BOOL(quit_option),
       ECORE_GETOPT_VALUE_BOOL(quit_option),
       ECORE_GETOPT_VALUE_BOOL(quit_option),
@@ -1110,8 +1113,6 @@ main(int argc, char *argv[])
 
    eina_init();
    ecore_init();
-   ecore_evas_init();
-   edje_init();
 
    if (argc < 2)
      {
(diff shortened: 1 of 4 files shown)


Coquinho:
I was thinking about `in-tree` shouldn't this meson change be inside `sys_windows` guards?

JPTIZ:
> 
> 
> I was thinking about `in-tree` shouldn't this meson change be inside `sys_windows` guards?

I don't see a reson. To be honest, depending on environment variables for this doesn't seem to be the best anyway.

joaoantoniocardoso:
> > I was thinking about `in-tree` shouldn't this meson change be inside `sys_windows` guards?
> 
> I don't see a reson. To be honest, depending on environment variables for this doesn't seem to be the best anyway.

I can't name any other viable way.

Coquinho:
> > I was thinking about `in-tree` shouldn't this meson change be inside `sys_windows` guards?
> 
> I don't see a reason. To be honest, depending on environment variables for this doesn't seem to be the best anyway.

And wich approach would be better?

JPTIZ:
> 
> 
> > > I was thinking about `in-tree` shouldn't this meson change be inside `sys_windows` guards?
> > 
> > 
> > I don't see a reason. To be honest, depending on environment variables for this doesn't seem to be the best anyway.
> 
> And wich approach would be better, @JPTIZ ?

This cmdline argument o/

JPTIZ:
> 
> 
> > > > > I was thinking about `in-tree` shouldn't this meson change be inside `sys_windows` guards?
> > > > 
> > > > 
> > > > I don't see a reason. To be honest, depending on environment variables for this doesn't seem to be the best anyway.
> > > 
> > > 
> > > And wich approach would be better, @JPTIZ ?
> > 
> > 
> > This cmdline argument o/
> 
> @JPTIZ, yeah, but we still need to pass it thought meson. And we don't need to pass it when building for linux.

We do! In Linux we set it through the environment variable. The whole point is that it must run inside EFL tree, why wouldn't we then set it in meson, since is 100% part of the build process?

Coquinho:
> > > > > > I was thinking about `in-tree` shouldn't this meson change be inside `sys_windows` guards?
> > > > > 
> > > > > 
> > > > > I don't see a reason. To be honest, depending on environment variables for this doesn't seem to be the best anyway.
> > > > 
> > > > 
> > > > And wich approach would be better, @JPTIZ ?
> > > 
> > > 
> > > This cmdline argument o/
> > 
> > 
> > @JPTIZ, yeah, but we still need to pass it thought meson. And we don't need to pass it when building for linux.
> 
> We do! In Linux we set it through the environment variable. The whole point is that it must run inside EFL tree, why wouldn't we then set it in meson, since is 100% part of the build process?

@JPTIZ, humm, ok, so at linux build `EFL_RUN_IN_TREE` is always setted to 1?

Coquinho:
> See: we use `edje_codegen_exe`, that is set [here](https://github.com/expertisesolutions/efl/blob/aaac9fdc754c4acaa534e4eb3bfe316833c2824b/src/bin/edje/meson.build#L105):
> 
> ```
>     edje_codegen_exe = [env, asan_option, 'EFL_RUN_IN_TREE=1', edje_codegen.full_path()]
> ```
> 
> Which will translate to...
> 
> ```
> > EFL_RUN_IN_TREE=1 edge_codegen.exe
> ```
> 
> Just happens that in Windows this has no effect. With this patch, `EFL_RUN_IN_TREE` could actually be discarded from `edge_codegen_exe` (maybe we should set `-t` in that line I liked above).

@JPTIZ, Oh, sure sure. Make sense. Thanks.

PR #393, opened 2020-08-19
Add evil_stdlib_mkstemp_create_many

joaoantoniocardoso:
While investigating temporary file creation from evil I noticed that there are no tests to check if the implementation of mkstemp can create several files in sequence. 

At the time I need to check that, than I wrote this simple test. 

Maybe it worth to add in our current branch, what do you think?


src/tests/evil/evil_test_stdlib.c
```diff
@@ -205,6 +205,34 @@ EFL_START_TEST(evil_stdlib_mkstemp)
 }
 EFL_END_TEST
 
+EFL_START_TEST(evil_stdlib_mkstemp_create_many)
+{
+   const unsigned long long files_to_create = 1000;
```
Coquinho, line 210: Why 1000?

src/tests/evil/evil_test_stdlib.c
```diff
@@ -205,6 +205,34 @@ EFL_START_TEST(evil_stdlib_mkstemp)
...
+   int fds[files_to_create];
+   const char template[] = "file_XXXXXX";
```
Coquinho, line 212: Maybe use `getenv("TEMP")` to create files at a temporary directory instead of build?

src/tests/evil/evil_test_stdlib.c
joaoantoniocardoso: First I thought to create as many files as possible, trying to reach the posix limit, but that is completely nonsense: it will take _forever_ to finish, possibly reaching timeout. 
Than I put some arbitrary quantity to be described as 'many'. :no_good:

Coquinho: welp...probably there are few cases where people create as many as 1k temporary files, sooooo, fair enough

src/tests/evil/evil_test_stdlib.c
joaoantoniocardoso: done

src/tests/evil/evil_test_stdlib.c
joaoantoniocardoso: Maybe it should be dynamically allocated?

JPTIZ: Also, any specific reason for `unsigned long long`? If it is to conform with array length type, just use an straight forward `size_t`.

src/tests/evil/evil_test_stdlib.c
```diff
@@ -205,6 +205,47 @@ EFL_START_TEST(evil_stdlib_mkstemp)
...
+        fail_if(fds[files_created] < 0);
+     }
+
+   // Close temporary files
+   for (files_created = 0; files_created < files_to_create; files_created++)
+     fail_if(close(fds[files_created]) == -1);
+
+   // Remove temporary files
+   for (files_created = 0; files_created < files_to_create; files_created++)
+     fail_if(unlink(templates[files_created]) == -1);
```
JPTIZ, line 244: Careful: this (declare without value to use inside one or multiple for-loops) is intensely error-prone (from accidental usage of unitialized value to accidental usage of an unexpected value that was given in a previous loop). I really would appreciate if it was used and declared in-loop:

```suggestion
   char templates[files_to_create][template_len];
   int fds[files_to_create] = { NULL };
   // Create temporary files
   for (size_t files_created = 0; files_created < files_to_create; files_created++)
     {
        strncpy_s(templates[files_created], template_len, template, template_len);

        fds[files_created] = mkstemp(templates[files_created]);
        fail_if(fds[files_created] < 0);
     }

   // Close temporary files
   for (size_t files_created = 0; files_created < files_to_create; files_created++)
     fail_if(close(fds[files_created]) == -1);

   // Remove temporary files
   for (size_t files_created = 0; files_created < files_to_create; files_created++)
     fail_if(unlink(templates[files_created]) == -1);
```

You don't have to worry about allocating multiple variables, since what will happen is that the compiler will reuse the same register for them, plus having additional guarantees over the expected variable initialization/value.

PR #396, opened 2020-08-20
Eet: Resolve eet_test_identity test case

joaoantoniocardoso:
Before:
```PowerShell
24/35 eet-suite                 TIMEOUT        37.92s

--- command ---
Running suite(s): Eet
95%: Checks: 22, Failures: 1, Errors: 0
../src/tests/eet/eet_test_identity.c:74:F:Eet Identity:eet_test_identity_simple:0: Failure '-1 == (fd = mkstemp(file))' occurred
```

After solving temporary file creation:
```PowerShell
Running suite(s): Eet
OPENSSL_Uplink(00007FF9E25D9D30,08): no OPENSSL_Applink
1/1 eet-suite FAIL           0.97s (exit status 1)
```

After including `openssl/applink.c`:
```PowerShell
Running suite(s): Eet
24/35 eet-suite                 OK             2.69s
```

### DIFF
--- a/src/tests/eet/eet_suite.c
+++ b/src/tests/eet/eet_suite.c
@@ -8,6 +8,14 @@
 #include "../efl_check.h"
 #include <Eet.h>
 
+#ifdef _MSC_VER
+/* On windows, applink.c needs to be included in the executable; otherwise
+ * all the tests which are using file pointer will fail in runtime, throwing
+ * `No OPENSSL_Applink` messages, i.e: eet_test_identity.
+ * */
+# include <openssl/applink.c>
+#endif
+
 char argv0[PATH_MAX];
 
 static const Efl_Test_Case etc[] = {
--- a/src/lib/eet/eet_cipher.c
+++ b/src/lib/eet/eet_cipher.c
@@ -32,6 +32,7 @@
 #  include <openssl/evp.h>
 #  include <openssl/x509.h>
 #  include <openssl/pem.h>
+#  include <openssl/bio.h>
 # endif /* ifdef HAVE_GNUTLS */
 #endif /* ifdef HAVE_SIGNATURE */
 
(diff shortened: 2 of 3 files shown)


src/tests/eet/eet_suite.c
```diff
@@ -8,6 +8,14 @@
 #include "../efl_check.h"
 #include <Eet.h>
 
+#ifdef _MSC_VER
+/* On windows, applink.c needs to be included in the executable; otherwise
+ * all the tests which are using file pointer will fail in runtime, throwing
+ * `No OPENSSL_Applink` messages, i.e: eet_test_identity.
+ * */
+# include <openssl/applink.c>
+#endif
```
Coquinho, line 17: Does it need to be specifically at `exe`? Could it be at a `dll`? I mean could we put it at `avil` guarded with `HAVE_OPENSSL` (which is defined at `meson.build`)

JPTIZ: > Does it need to be specifically at `exe`?

Yes, [as stated in OpenSSL FAQ](https://www.openssl.org/docs/faq.html#PROG3).

> I mean could we put it at `evil` guarded with `HAVE_OPENSSL` (which is defined at `meson.build`)

An alternative could be make a `INCLUDE_OPENSSL_APPLINK` symbol that is only true when compiling the executable, but to be honest if we would do this, I prefer to just ask to include `<openssl/applink.c>`.

Coquinho:
And, with this `eet-suite` works:
```
24/35 eet-suite                 OK             2.69s
```

src/tests/eet/eet_suite.c
```diff
@@ -8,6 +8,14 @@
 #include "../efl_check.h"
 #include <Eet.h>
 
+#ifdef _MSC_VER
+/* On windows, applink.c needs to be included in the executable; otherwise
+ * all the tests which are using file pointer will fail in runtime, throwing
+ * `No OPENSSL_Applink` messages, i.e: eet_test_identity.
+ * */
+# include <openssl/applink.c>
```
JPTIZ, line 16: Opened the code crossing my fingers that we didn't need applink.

Guess I was wrong. Oh well!

joaoantoniocardoso: hahah yeah, I tried my best with no success removing it...

PR #400, opened 2020-08-25
Use eina_file_unlink() instead of unlink()

joaoantoniocardoso:
As `tmpfile` are being created using `eina_file_mkstemp()`, we can use `eina_file_unlink()` instead of `unlink()`, which closes the handles before unlink (see implementation [here](https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/src/lib/eina/eina_file_win32.c#L794https://github.com/expertisesolutions/efl/blob/devs/expertise/native-windows/src/lib/eina/eina_file_win32.c#L794)).
This also removes the need to call `eet_clearcache()` before `unlink()` as the handles are being closed by `eina_file_unlink()`.

This patch is a candidate to solve https://phab.enlightenment.org/T8795.

### DIFF
--- a/src/examples/eet/eet-data-cipher_decipher.c
+++ b/src/examples/eet/eet-data-cipher_decipher.c
@@ -111,7 +111,7 @@ main(void)
    eet_close(ef);
 
 error:
-   if (unlink(file) != 0)
+   if (eina_file_unlink(file) != EINA_TRUE)
      {
         fprintf(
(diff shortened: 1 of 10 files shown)


joaoantoniocardoso:
> Sounds like a good patch to submit to phab.

Yeah, we can send it together with the patch that changes to use `eina_file_mkstemp()` instad of `mkstemp()`.

PR #402, opened 2020-08-26
eina: Resolve eina_file01.exe

Coquinho:
Use `eina_environment_home_get` to get home dir instead of using
`/home/`

### DIFF
--- a/src/examples/eina/eina_file_01.c
+++ b/src/examples/eina/eina_file_01.c
@@ -16,7 +16,7 @@ main(int argc EINA_UNUSED, char **argv EINA_UNUSED)
    Eina_Iterator *it;
    const char *f_name;
    const Eina_File_Direct_Info *f_info;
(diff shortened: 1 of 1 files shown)


src/examples/eina/eina_file_01.c
```diff
@@ -16,25 +16,27 @@ main(int argc EINA_UNUSED, char **argv EINA_UNUSED)
    Eina_Iterator *it;
    const char *f_name;
    const Eina_File_Direct_Info *f_info;
+   const char* homedir = eina_environment_home_get();
```
joaoantoniocardoso, line 19:
```suggestion
   const char *homedir = eina_environment_home_get();
```

Coquinho: Sure, why not... but why?

joaoantoniocardoso: To me, it makes easier to read.

PR #405, opened 2020-08-28
Fix: add libcurl to meson

joaoantoniocardoso:
libcurl cannot be used if meson doesn't link it:
```PowerShell
ERR:ecore_con ../src/lib/ecore_con/ecore_con_url_curl.c:255 _c_init()
Cannot find libcurl at runtime!
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -91,6 +91,9 @@ jobs:
     - name: Install getopt (vcpkg)
       if: steps.cache-vcpkg.outputs.cache-hit != 'true'
       run: .\vcpkg\vcpkg install getopt:x64-windows
+    - name: Install curl (vcpkg)
+      if: steps.cache-vcpkg.outputs.cache-hit != 'true'
+      run: .\vcpkg\vcpkg install curl:x64-windows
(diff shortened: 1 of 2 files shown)


.github/workflows/windows.yml
```diff
@@ -97,6 +97,9 @@ jobs:
     - name: Install getopt (vcpkg)
       if: steps.cache-vcpkg.outputs.cache-hit != 'true'
       run: .\vcpkg\vcpkg install getopt:x64-windows
+    - name: Install curl (vcpkg)
+      if: steps.cache-vcpkg.outputs.cache-hit != 'true'
+      run: .\vcpkg\vcpkg install curl:x64-windows
```
Coquinho, line 102: woops, you need to run it once without checking for the cache, as curl is not cached, and them update the PR adding the verification.

joaoantoniocardoso:
Thanks @JPTIZ for fixing the cache thing, it is working now =)
@JPTIZ  and @Coquinho, could you take a look again?

.github/workflows/windows.yml
joaoantoniocardoso: Solved in #409

PR #409, opened 2020-09-01
ci: Make vcpkg cache unique for each workflow change.

JPTIZ:
In order to have better control over dependencies on vcpkg on CI, this PR uses a hash of the current contents of `.github/workflows/windows.yml` as the vcpkg cache key. This means that vcpkg cache will only be invalidated (i.e. vcpkg will be cloned & setuped and then dependencies will be installed one by one) when that file changes (possibly by adding a new dependency).


JPTIZ:
Fixed vcpkg key for `tests:`.

PR #413, opened 2020-09-02
Fix usage of temporary dir (/tmp/) [for the remaining]

joaoantoniocardoso:


JPTIZ:
I really would appreciate if you could be more specific in PR names :) Prefixing with `lib: ` could be a good start.

JPTIZ:
Oh I see you just changed the name to clarify

src/examples/evas/evas-images3.c
```diff
@@ -30,10 +30,10 @@ static const char *commands = \
   "commands are:\n"
   "\tw - write new pixel data to image\n"
   "\ti - print image info\n"
-  "\ta - save noise image to disk (/tmp dir)\n"
+  "\ta - save noise image to disk (TEMP dir)\n"
   "\th - print help\n";
 
-const char *file_path = "/tmp/evas-images2-example.png";
+const char *filename = "evas-images2-example.png";
```
Coquinho, line 36: This one is kind strange. It don't seems to be used anywhere.

joaoantoniocardoso: Yeah, indeed it isn't used anywhere, but as it is a copy of the same template of its brothers `evas-image[2345].c` I changed here too.

JPTIZ (approved):

PR #414, opened 2020-09-02
Use eina_file_unlink() instead of unlink()

joaoantoniocardoso:
The `unlink()` function is being used all over the place where `eina_file_unlink()` would possibly fit better.

Note that here I first used a macro to perform the work, so I am still analyzing if it makes sense. For example in `src/bin/edje` there are several uses of `unlink()` that cannot be directly changed to `eina_file_unlink()` or it won't compile, so I didn't add to this batch.

### DIFF
--- a/src/bin/efreet/efreet_mime_cache_create.c
+++ b/src/bin/efreet/efreet_mime_cache_create.c
@@ -373,7 +373,7 @@ write_error:
    ERR("Cannot write to %s", buf);
 error:
    if (f) fclose(f);
-   if (unlink(buf) != 0) WRN("Cannot delete tmp file %s", buf);
+   if (eina_file_unlink(buf) != 0) WRN("Cannot delete tmp file %s", buf);
 }
 
 int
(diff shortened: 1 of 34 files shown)


src/bin/efreet/efreet_mime_cache_create.c
```diff
@@ -373,7 +373,7 @@ store_cache(const char *out)
    ERR("Cannot write to %s", buf);
 error:
    if (f) fclose(f);
-   if (unlink(buf) != 0) WRN("Cannot delete tmp file %s", buf);
+   if (eina_file_unlink(buf) != 0) WRN("Cannot delete tmp file %s", buf);
```
JPTIZ, line 376: Probably you'll want to change to `!= EINA_TRUE`.

joaoantoniocardoso:
```suggestion
   if (eina_file_unlink(buf) != EINA_TRUE) WRN("Cannot delete tmp file %s", buf);
```

PR #415, opened 2020-09-03
 ecore-evas: Fix load of engines modules for native-windows

Coquinho:
Change the search path of `ecore-evas` engines modules for native-windows so it uses the `PACKAGE_LIB_DIR`

Test
------
`exemple/elementary/label_example_01.exe` should open a window as opesed to before this PR.


.gitignore
```diff
@@ -96,6 +96,9 @@ efl_libs.csv
 .vscode/
 .uuid
 
+# My ignores
+prefix/*
+
```
JPTIZ, line 101: If it is local, you can add this to `.git/info/exclude`.

joaoantoniocardoso (approved):

Coquinho:
This isn't needed.
I was wrongly using `build/**/*.dll` instead of `<installed dir>/bin`, which raised the need for this PR.

PR #418, opened 2020-09-08
win32: Font search

felipealmeida:


felipealmeida:
I pushed because they needed to test in Korea

PR #449, opened 2020-09-15
elementary_test: Fix division by zero

### LINKED ISSUE #446: elementary_tests: `Genlist Tree` button crashes, closing everything.
## Current behavior
On Windows, clicking on the button `Genlist Tree` in the `Genlist Item Show` example from `elementary_test` crashes, closing everything, even the elementary_test window, and the cmd window.

## Expected behavior
A window with some content should be created by clicking on the button `Genlist Tree`.

## Steps to reproduce
1. Run `elementary_tests.exe`
2. Open `Genlist Item Show`
3. Click `Genlist Tree`

felipealmeida:
Fixes #446

### DIFF
--- a/src/bin/elementary/test_genlist.c
+++ b/src/bin/elementary/test_genlist.c
@@ -5587,7 +5587,7 @@ test_genlist_show_tree(void *data,
    else max = 2000;
    for (i = 0; i <= max; i++)
      {
-       if (i % (max / 20) == 0)
+       if (i % ((max / 20) ? 0 : 1) == 0)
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso (approved):
~While this patch does fixes the error of crash when no numbers are set, the Windows crash when clicking `Genlist Tree` remains, so this patch is in fact interesting but it doesn't solve #446.~

edit: It does fixes 446, except that when you abuse setting a big tree like 100, it crashes.

PR #470, opened 2020-09-23
devs/jptiz/lua optional

JPTIZ:
- ci: travis: switch exactness testing to normal test file
- exactness docs: Add docs, fix docs, remove docs
- comments: Typos in Efl_Config.h
- doxygen docs: Add _Eina_Version struct docs
- doxygen docs: some fixes in Evas_Legacy.h
- free memory in clipboard notify event and fix clipboard clear event
- edje_legacy: Remove EINA_UNUSED when obj is actually used
- code: Silence compiler warnings.
- elm_entry: using keycode for main keyboard shortcuts
- ecore-evas/cocoa: fix cnp leak
- evas/text: remove paragraph_direction inherit during render
- Replace strncmp code
- eo: use eina_prefetch
- efl_ui/layout: correctly batch object changed state when using freeze/thaw
- tests/elm: add helper macros for asserting size and position of objects
- tests/elm: add wrapper for forcing manual render in tests
- tests/layout: add explicit test for freeze and thaw no-op correctness
- tests/hover: add behavior tests for elm_hover
- ecore_evas: fix leaks in macos and win32
- emile: fix a potentional resource leaking.
- build: add version to exactness libs
- ui textpath: update map before rendering begins.
- ui textpath: removed unused var.
- efl.ui.textbox: add test case for copy/paste results
- elm icon/image efl ui image - respect aspect hints at all if set
- elm icon/image efl ui image - respect aspect hints at all if set
- eo: only prepare a empty node
- evas: prepare for cur beeing NULL
- ci: force system zlib on osx
- meson: support unified suffixes for modules
- meson: build modules on macos as .so
- meson: do no build upower on macos
- ecore_evas_cocoa: be aware of NULL data
- efl_ui_dnd: cleanup when backend failed to create dnd op
- elm_entry, accessibility: Send utf8 text instead of markup to ATSPI client
- release: Update NEWS and bump version for 1.24.0-beta1 release
- elm_dnd: null check for safety
- tests: Add case to match ephoto behaviour.
- tests: popup
- efreet - try less chaotic efreetd restart and delay 0.5-1.0 rand sec
- efreetd - ensure on bsd where connects succed despite daemon being gone
- evas: fix dereference after null check
- ecore-evas - x dnd/cnp - handle null data content and not segv
- ecore con - clean up file lock andunlock on failed bind
- fix buid on bsd
- edje_calc: add null check after eet_read
- elm_atspi_bridge: fix memory leak
- Ector: Refer to visibility of Ector_Renderer when doing rle_update.
- evas text: fix memory leak.
- elm_cnp: fix memory leaks
- eo: null check for vtable allocation
- evas_main: add null check for safety
- Efl.Canvas.Vg.Container: Prevent duplicate transpacency calculation.
- efl_text_formatter: rename from Efl.Text.Formatter to Efl.Text_Formatter
- Fix namespace clashes for Efl.Text.Cursor
- efl_input_text: rename namespace
- meson: do not install namespace problem legacy files
- efl/wl: fix namespacing conflict in rotation enum
- efl_ui_focus_manager: refactor _request_subchild_except
- efl_ui_focus_manager: do not fallback to the same object
- efl_ui_focus_manager: do not magically set focus to new elements
- efl_ui_focus_layer: correctly dismantle redirect chain
- elm_slider: create timer for hiding later
- ecore_evas_x: allow parsing of links form the link list
- elm_dnd: support target setting on none evas_objects
- ecore_evas: add API for finding out if event is used
- ecore_evas_x: reject the promise instead of return
- ecore_evas_x: add safety check for data in converter
- build: fix libraries of efreet-mime -trash
- elm_dnd: complete the list of text types
- ecore_evas_x: fix typo'ed function name
- efl_ui_focus_manager: correct check
- doxygen docs: Fix double group
- doxygen docs: Proper usage of @ingroup and @addtogroup
- doxygen docs: Fix misplaced Evas_Font group tags
- doxygen docs: Fix missing docs in Edje
- doxygen docs: Fix several invalid refs in Elm
- doxygen docs: Fix Elm_Toolbar groups
- doxygen docs: Define a couple missing ELM_WIN types
- doxygen docs: Fix invalid doxygen output
- release: Update NEWS and bump version for 1.24.0-beta2 release
- elm menu - fix segv on deletion/shutdown where parent was not done ok
- tests/efl: use alloca for composite model test future to avoid leak
- ecore_cocoa: fix compile error using 10.15 SDK
- evas: text_cursor: check return value and print error
- evas_common_font: prevent leaking RGBA_Font
- Revert "evas: text_cursor: check return value and print error"
- toolbar - dont create icon objects then del if icon is null
- evas_textblock : fix text insertion & selection with ps in single line
- efreet - look in /usr/local/share/pixmaps too
- efreet - also look in lib in /usr/local/share/pixmaps too
- efreet - disable strict as this makes things less functional...
- tests/elm: add callback tests for elm_list
- elm/hoversel: use a wref to accurately track internal hover object
- elm/hoversel: remove unnecessary internal callback deletion
- tests/elm: add hoversel behavior tests
- elm/hoversel: force calc on internal hover object during activate
- tests/elm: add hoversel test to verify edge of canvas positioning
- tests/elm: explicitly destroy hoversel while active
- ecore_evas: fix fallback code
- evas_test: change START_TEST to EFL_START_TEST
- evas_textblock: pick textrun fonts
- evas_textblock: remove special treat for variation sequence text run
- exactness: remove no longer used code since move to preload
- exactness: fix typo in function name
- exactness: limit length for font_path to stay in buffer
- elm/win: declare prototype for ecore_evas_dnd_mark_motion_used
- eina_content: free eina content in the test:
- eina/value: use correct time types when comparing time types
- evas_textblock: prevent segfault if there are no style
- canvas svg: fix memory leak.
- tests/elm: add image tests to verify internal object size is clamped to object size
- edje: fix cursor scrolling
- evas cache - up padding just in case i was off by one
- ecore evas - wl - fix useless iff case for rotation
- ecore - proc env - fix values return NULL case - coverity
- ecore - examples - fix gnutls fd handler example warnings
- elm - genlist test - handl g_data being null
- exactness-inspect: sanitize if clause
- exactness-inspect: fix NULL access
- exactness-inspect: unit1 can never be NULL
- Revert "evas_textblock: pick textrun fonts"
- Fixing POTFILES and updating italian translation
- evas render - flush, dont clear unref/clean arrays on  render done
- evas_textgrid: update text props to use current script font
- evas_textblock/evas_text/evas_textgrid: pick textrun fonts
- eolian: add composites into hard deps when checking
- remove unused imports in eo/eot files
- eolian: globally enable eolian_state_check, except for beta classes
- Updating all po files
- elm - theme - add outline style for frames
- elm theme - add support for smooth restart for e
- exactness: remove unused code
- exactness: handle case where eet_data_image_write() fails
- exactness: make sure we handle a negative file descriptor
- exactness: make sure we handle a negative file descriptor
- doxygen docs: Fix all invalid @param names
- exactness-inspect: remove abscure xor check
- exactness-inspect: do not use sprintf
- eolian: only warn instead of erroring on unused imports
- eolian: fix C type serialization of error()
- ecore-evas-wayland: Fix formatting
- ecore-drm2: Change ERR to WRN
- evas_object_textblock: add null check for return of eina_inlist_remove
- release: Update NEWS and bump version for 1.24.0-beta3 release
- meson: do not install .eo files of not installed header files
- elementary: do not install legacy .eo files
- eo: install eot files correctly
- evas: remove dead struct definition
- elementary: these definitions are in real header files
- eina: add new definitions to a old enum
- elementary: fix string definition
- eolian: clarify @move tag usage in documentation
- ecore_wl2: EFL Wayland Build on FreeBSD
- ecore_file - address issue about false positive of finding a file in ./
- EFl_Canvas_Text: add typedef for backward compatibility in legacy headers
- Revert "meson: do not install .eo files of not installed header files"
- exactness: do not access potential NULL pointer
- exactness: use convinience ecore_file function to get dir
- exactness: check action for NULL
- evas_gl_context: reduce conditions for image orientation check
- canvas layer: find the top item for only valids.
- canvas layer: +prevent null layer objects.
- evas gl - shader cache was needless losing and overwriting
- evas gl - fix coverity complaint in shader cache update
- efreet - work around conconforming apps putting startupwmclass in quotes
- eolian: require opaque structs to be used by ref always
- elm - combobox - pretty borken but make it a little less
- doxygen docs: Document compositor sync API as deprecated
- efl_ui_image: Fix unnecessary unload of memfile_set.
- doxygen docs: Put APIs in correct group
- doxygen docs: Match doc with correct enum for Render_Op
- doxygen docs: Add missing Elementary group
- doxygen docs: Fix invalid refs to Layout
- doxygen docs: Put Evas_Canvas legacy methods in the right group
- doxygen docs: Fix most errors on Evas_Legacy.h
- elm/cnp: remove const from static function return
- canvas clip: replace hash to cached array.
- efl_ui_focus_manager_calc: correctly flag elements used
- efl_fileselector_entry: do not crash on a invalid model
- elm/list: avoid erroring on non-elm item contents
- tests/elm: add a list sizing test
- tests/elm: check all objects for signal interface before calling signal funcs
- tests/elm: add some more specific convenience macros for size checks
- tests/elm: add util callback for capturing event_info in smart callbacks
- tests/elm: add more naviframe tests
- meson: allow debug-threads to be disabled in debug
- build: install eo files per default
- ecore-evas-drm: Fix broken init count
- ecore-evas: Fix formatting
- ecore-drm2: Change _output_crtc_find to accept an int
- eolian_gen: hack around function pointers generating unused params
- eolian_gen: mark empty setter value params as unused
- canvas vg: Correct order of data free.
- doxygen docs: Move CnP API out of the Remove Canvas group
- doxygen docs: Add @since tags to EE selection API
- release: Update NEWS and bump version for 1.24.0 release
- build: release freeeze over
- edje: efl: add new EFL_VERSION_1_25 define for new cycle
- doxygen docs: Fix broken links in main page
- doxygen docs: Fix missing links in Ecore_Drm
- doxygen docs: Replace broken URL with doxygen ref
- doxygen docs: Fix reference to Elm_Start
- doxygen docs: add missing example references
- doxygen docs: Add tests folder to EXAMPLE_PATH
- doxygen docs: Add missing group braces
- elm_fileselector: support expandable items
- ecore_evas: only define these eina errors once
- exactness: remove no longer needed _evas_new()
- build - remove useless header checks for uv.h
- edje - don't set font sizes less than 1
- ecore-drm2: Add support for using page_flip_handler2
- ecore-buffer: Fix formatting
- ecore_evas_x - fix maximize/unmaximize request corner case
- examples: cxx: remove forced warning on unfinished APIs
- remove arguments of LOGFN in ecre_x and ecore_wayland as the are always the same
- evas_textblock: optimize calculate main format once in layout setup stage
- exactness_player: do not test for delay_ms
- ecore_drm: ensure we have the symbol available at the right place.
- ci: travis: update first few jobs to Fedora32
- tests: edje: Fedora32 system fail resolve
- ci: travis: update more jobs to Fedora32 image
- evas_textblock: clear paragraphs during markup_set
- edje_textblock: optimize priority handling for content_fit options
- edje_textblock: content_fit size_range update
- travis: update asan build to Fedora 32
- ci: travis: switch coverity build back to Fedora 31 image
- remove arguments of LOGFN in  ecore_wayland modules as the are always the same
- ecore-evas-drm: Remove useless check for visibility
- canvas map: draw direct image as possible.
- evas: fix png regression issue
- cxx: Initialize _list and _node with nullptr
- Use __func__ C99 identifier instead of __FUNCTION__ compiler extension
- tests: ecore_con: renew test ssl certificate which expired in 2016
- ci: travis: update Debian to 10.3 and Ubuntu to 20.04
- tests/elm: add callback testing to image memfile test
- evas: use SEQUENTIAL for decoding png file
- doxygen docs: Complete docs for Eina_Rectangle
- doxygen docs: Complete docs for Eina CPU functions
- doxygen docs: Complete docs for Eina Log
- doxygen docs: Complete docs for Eina Module
- doxygen docs: Fix grouping mess in Eina Simple XML
- doxygen docs: Complete docs for Eina Thread
- doxygen docs: Put Eina XAttrs in their own group
- doxygen docs: Fix problems in Evas_Canvas
- eina: resolve float comparison warnings
- edje_cc: fix comparisons for possible non-float values
- edje_cc: consistently use eina macros for all float comparisons
- tests: resolve float comparison warnings
- bin/elm: resolve float compare warnings
- elm: resolve float comparison warnings
- evas/events: fix conditional for emitting source mouse move events
- evas + gl - fix callback-to-app-changes-state issue with pixel data
- canvas render: don't apply mask if the objects in the map.
- tests: ecore_con another cert update to extend the time its valid to 100 years
- efl_canvas_vg_node: fix the implemenation function of the efl_canvas_vg_node(evas_object_above_get, evas_object_below_get)
- evas_vg_node: deprecate evas_vg_node_geometry_set.
- evas_object_textblock: fix to free style properly by edje's style_user_pop
- build: ensure we set ENABLE_LUA_OLD when using a lua interpreter
- evas_object_textblock: reduce layout calculations
- Efl.Canvas.Object.Animation: Prevent crash when called .animation_pause_set.
- evas: resolve float compare warnings
- evas/filter: resolve float compare warnings
- evas/gl: resolve float compare warnings
- examples: resolve float compare warnings
- exactness: resolve float compare warnings
- edje: resolve float compare warnings
- static/triangulator: resolve float compare warnings
- build: enable -Wfloat-compare by default for dev builds
- build: fix void* use in pointer maths
- build: enable -Wpointer-arith for dev builds
- Efl.Ui.Button_Legacy: Prevent duplicate signal emit of contents
- Efl.Ui.Progressbar_Legacy: Prevent unnecessary signal emit for contents
- Efl.Ui.Check_Legacy: Prevent unnecessary signal emit for contents
- Efl.Ui.Radio_Legacy: Prevent unnecessary signal emit for contents
- Revert "canvas render: don't apply mask if the objects in the map."
- edje_calc: Add unload file for already loaded vector lottie file.
- elm: refactor legacy _icon_signal_emit
- ci: travis: switch exactness build to Fedora32 image
- evas_vg_shape: return NULL in evas_vg_shape_add when the parent of shape is NULL
- ecore-drm2: Add API function to return default depth
- ecore-drm2: Add missing params for doxygen
- efl_ui_image: fix to call "clicked" smart callback when there is no img
- systemd - make libsystemd use/supprot entirely runtime "dlopened"
- Fix EAPI definition by defining EFL_BUILD for each built DLL
- ecore - systemd - missed an ifdef for sytemd disable
- elgind - it's apparently libelogind-shared ...
- Revert "Fix EAPI definition by defining EFL_BUILD for each built DLL"
- ecore-evas-wayland: Fix formatting according to coding conventions
- ecore-evas-drm: Remove duplicate config set in meson.build file
- elput: Fix formatting
- elput: Rmove empty pub_eo_file_target in meson.build
- png: handle file name .9.png
- ecore_anim: rework bezier curve function
- evas: do not call evas_object_change in reneder_post
- ecore_con: Fix unused argument warning
- tests: eina: fix usage of long on Windows
- evas: gesture_manager: fix usage of long on Windows
- ecore-x - ensure we fully look at all xkb events and refresh all binds
- efl_ui_layout: remove group_calcualte to respect legacy min size
- meson: Changed meson build to support address sanitizer build
- elm_genlist/elm_gengrid: fix memory leak caused by efl_wref_del.
- elua: remove old lua bindings generator
- elua: prepare for compatibility with cffi-lua (use ffi.nullptr)
- evas_vg_container: return NULL in evas_vg_container_add when the parent of container is NULL
- ecore - expose "hidden" internal api for debugging to get wakeup time
- ecore-x expose some debug times like the exact time the vsync woke
- ecore thread - feedback threads should not be background threads...
- ecore-x : move x vsync device back to mainloop from thread
- ecore evas - tick printf debugs - disabled right now so no noise
- ecore - another systemd header - go away. don't need it now.
- ecore_evas: fix uninitialized variable warning
- elm test: include evil_private.h to declare realpath() on Windows
- evas image: reload after alpha_set
- elm_dnd: do not register widgets twice
- ecore_evas_wayland_common: keep the backword compatibility for wayland dnd handler
- eina - dont use SCHED_BATCH or SCHED_IDLE unless they are defined
- build: set DLL_EXPORT when on a windows system
- benchmark: fix ident in meson.build file
- libs: headers: ensure we include headers in the correct order for EAPI
- edje: fix animation memory leak.
- canvas vg: fix dangling vector containers.
- build: move EFL_BUILD to package_c_args used in all subprojects
- benchmarks: use EFL_BUILD=1 in c_args
- build: lib: harmonize the use of package_c_args in all libs
- build: modules: harmonize use of package_c_args
- lib: add more c_args handling for places we missed before
- Merge efreet_mime efreet_trash into efreet
- evas vg: clean up dangling instances
- canvas vg: fix vector file data memory leak.
- evas - csd and wl csd - fix alpha zeroing to avoid framespace junk
- ecore_main_timechanges: fix include
- build: disable leak detection in address sanitizer build for tests as well
- evas_textblock: fix doc typo
- ecore_wl2_window: do not unmap subsurface in hide
- ecore_wl2 : Organize window creation/deletion events.
- ecore-wl2: Fix build break from patch 9ca13ef5c10904b620407
- ecore-wl2: Remove unused variables
- ecore-evas-wayland: Fix unused variables
- edje: Fix memory leak by freeing the filter data.
- edje_load : clean up vector resource when edje file freed.
- edje: fix memory leak by freeing the limits pointer.
- efl gfx_path: fix memory leak.
- build: split up building of libs, bins, modules, benchmark
- refactor build
- build: time for efl-one
- build: stop buildsystem from beeing a public dependency
- build - fix deps for efl_canvas_wl to be able to build against efl
- build: add correct flags to efl-one.pc
- build - systemd service files - install fix if systemd not found
- Eina file test: on Windows, open cmd, not a shell
- Eina file test: use the correct Eina file object
- edje: fix memory leak in edje part recalculation.
- build: efl-one track all subdirs correctly
- ecore_wl2: Fix event type name.
- ecore_wl2_window: copy available rotation info.
- elput - it seems not everyone has major and minor macros defined
- elput - fix #else path for minor
- ci: travis: switch to newly prepared Fedora32 mingw image
- evas_render: do not use mask of proxy source
- efl_ui_bg: disable preloading of the internal image
- build: rely on automatic pkg file generation for eina
- build: add priv eo targets to efl_one_eo_deps
- build: make eina drag in the -lm flag
- build: split off static parts from external parts in evas
- build: vg_common depends on evas internals
- Introduce a test to verify efl-one only links to the correct libs
- build: add efl-canvasl-wl to efl-one
- build - fix eina pc deps so you can build against efl again
- edje edje_cc: remove beta tag for vector part.
- efl_ui_textpath: reduce unneeded matrix calculations
- Support WebP Animation Image Files
- evas_text: reduce Video Memory & RAM needed for text textures (Color Glyphs)
- build: ensure we set all needed asan option for check run as well
- ci: get rid of all extra asan handling
- elua: allow building with interpreted (non-luajit) lua
- bindings/luajit/eolian: get rid of using bitops
- bindings: rename luajit -> lua
- elua: fix build with luajit or lua 5.1
- elua: use cffi-lua public header
- evas - examples - evas box fix path to png file
- evas examples - build box example again - missing header
- bindings/lua: minor fixes to accommodate cffi-lua
- eolian: add out-param variants of expr eval/value get funcs
- bindings/lua: use new _fill api variants for expr values
- elua: fix object system on lua 5.2 onwards
- bindings/lua: fixes for cffi-lua
- bindings/lua: simplify some eolian binding code
- elm_gen****: free item_content map
- Efl.Gfx.Vg.Value_Provider: Move internal function changed_flag to eo.
- svg/loader: Fix memory leak
- elementary: Fix memory leak in elm_list
- elm_hoversel: Add missing resize_job's free
- canvas svg: fix missing break.
- eo: Fix memory leak in efl_key_wref_set() api.
- elm_gesture_layer: Arrange the logic for delete the target object in gesture cb.
- efl_access_text: remove ptr usage in eo files
- build: do not link test suite again against static libs
- ci: fix for rename luajit to lua bindings
- build: add eet to evas suite
- ci: travis: enable efl-one in our all enabled build
- ecore: Fix Asan stack overflow warning
- Efl.Canvas.Vg.Object: Optimize Ector Surface Size
- elementary_test: Added --autoclose option to elementary_test
- evas/engine: Fix memory leak in generic cache.
- Efl.Canvas.Vg.Object: Prevent null pointer access
- Revert "elementary: build elementary_test additionally as shared object"
- ci: remove not in use script
- ci: remove outdated README file
- ci: remove no longer used address sanitizer leak suppression file
- ci: log time in every fold start and end
- Efl.Gfx.Path: Modify boundary calculation.
- Efl.Canvas.Vg.Object: Fix backup size
- build: move static out of none_static
- elm test - animated icon - big win to have a bg and not garbage
- evas gl - tune gl engine to use a bit less cpu overhead
- need to upgrade meson. it looks like older ones have trouble
- tiler - add maxreg for newtiler - not used atm but matches current
- evas render - add debug to draw boxes around update regions
- evas - genetic update regions smart merge - merge v adjacent regions
- edje: markup flag should be TRUE in legacy edje_object_part_text_set
- evas: correctly declare filter_internal as public API
- build: do not publish surface .eo files
- emotion: remove all left overs from removed backends
- build: disable elua by default, plus nicer detection
- ci: drop lua from default binding set as elua is off by default
- eldbus: fix the data size issue.
- webp: add version check in configuration time
- ci: disable wepb loader until we have a change to update the mingw cross env
- ci: simplify ccache setup handling
- ci: travis: add more fold and timing markers in some forgotten places
- exactness: record and print test time on success
- canvas gl: remove out-dated compiler keywords
- ecore_ipc: Fix memory leak
- elementary/example: Fix memory leak in gengrid test
- evas_common_font: release reallocated glyphs bitmaps data
- evas_vg_cache: Don't caching vg file if value_provider is applied.
- doxygen docs: Fix API reference group to Ecore_Eo
- Revert "ci: disable wepb loader until we have a change to update the mingw cross env"
- eo: make the parent a optional constructor
- dotnet: Ignore Efl.Object.parent as constructor
- eo-files: convert all Eina.Strbuf & Eina.Binbuf
- cxx: Fix uses of intrinsic eolian binbuf type
- efl_ui_win: declare constructors not optional
- eina - abstratc content - smaller stack buffer for limited size string
- eina - prefix - use smaller buffers for building strings
- efreet - handle runtime relocation right with default XDG_DATA_DIRS
- ci: travis: disable the webp loader on CI builds for now
- mono: update window construction
- efl_ui_textpath: add the efl_ui_textpath_text_user_style_set API for applying the user text style
- ui textpath: updated doc.
- Use extern after EAPI
- tests: evas: check fd being non -1 before using close()
- tests: eina: check return of eina_module_load and unload
- evas stack: ++safe code.
- ecore_wl2_tests: Fix formatting
- tests/ecore_wl2: Fix unchecked return value
- tests/ecore_wl2: Fix unchecked return value
- ecore_drm: check return of drmModeSetCrtc()
- tests: eina: also handle failure cases when cleaning up created files/folders
- tests: eo: handle error return of eina_value_int_convert()
- tests: eina: make sure we always check on eina_matrixsparse_data_idx_get return
- efl - build again - test suite break in using undefined ERR macro
- eina: only enable EINA_LOG_BACKTRACE when backtrace API is present
- ecore: use standard LC_ALL instead of __LC_ALL in systemd module
- elm_calendar: Do not pass NULL to function.
- edje_cc_handlers: update documentation of "fixed" to be clearer
- modules: only build gl_drm evas engine when using es-egl
- tests: elm: check return of efl_config_bool_set
- efl_ui_image: fixed elm_image_preload_disabled_set api is not working when it is called before file set
- ui image: fix a missing else case.
- build: decrease the minimal needed meson verion to 0.50
- efl_ui_win: reuse bg which was created for version check
- modules: vnc: allow building for FreeBSD with a different header path
- meson_options: fix typo in option description
- meson_options: really fix the typo
- evas/engines/gl_generic: fix byte order after glReadPixels on BE
- webp - 0.5.0 is actually new enough
- modules: gl_x11: ensure we check the result of  evas_eglMakeCurrent()
- tests: edje: check result of evas_engine_info_set()
- Efl.Canvas.Vg.Object: Fix wrong render area
- evas_object_table: fix not to use hint_fill on elm_table
- Revert "ci: travis: disable the webp loader on CI builds for now"
- modules: eina: chained_pool: check for NULL before passing it to deref
- elementary/perf:remove unnecessary elm_widget_is() check
- elementary/perf: reduce number of efl_isa() call. in elementary test application efl_isa() shows up as hotsport during profiling so refactored the code to reduce redundant efl_isa() calls.
- efl_ui_widget: addition to the previous commit
- gl common - actually allow vertex limits
- evas gl common - remove comment - not valid anymore
- tests: eio: make sure we check return value
- elm_image: keep backword compatibility for elm_image_file_set API when setting url file set twice
- benchmarks: eina: make sure we do not divide by zero
- elm_image: remove the spaces and keep the indentation for elm_image_file_set()
- efreetd - cache - add more statinfo work around 0 mtime distros
- efreet - mtime 0 fix build on windows
- tests: eina: fix ENOMEM message on Windows and solaris
- meson: add Solaris support
- efreet caceh - fix coverity report
- elm_scroller: use region of proxy soruce for loop
- eina_unicode: have explicit type conversions
- Textblock : Fix cursor cluster movement when emoji at the line start
- eina_strbuf: resolve segfault when replace used with read_only buffer
- evas textblock : update font source when it set using font_source_set
- eina_strbuf_manage_new: update documentation
- eina_strbuf: if readonly strbuf is malloc, then it will stop being readonly
- elementary_test: remove leaked style
- eina_strbuf: introduce change last occurrence function
- Efl.Gfx.Path: Change draw of a rounded rect from arc to quadratic
- Get rid of trailing whitespace (1 / 14)
- Get rid of trailing whitespaces (2 / 14)
- Get rid of trailing whitespaces (3 / 14)
- Get rid of trailing whitespaces (4 / 14)
- Get rid of trailing whitespaces (5 / 14)
- Get rid of trailing whitespaces (6 / 14)
- Get rid of trailing whitespaces (7 / 14)
- Get rid of trailing whitespaces (8 / 14)
- Get rid of trailing whitespaces (9 / 14)
- Get rid of trailing whitespaces (10 / 14)
- Get rid of trailing whitespaces (11 / 14)
- Get rid of trailing whitespaces (12 / 14)
- Get rid of trailing whitespaces (13 / 14)
- Get rid of trailing whitespaces (14 / 14)
- eina_test_file.c: Do not test errno if the function doesn't fail
- ecore wl - make comment note about allocating buffers
- ecore-x vysnc - env var - fix typo
- efl_check.h: Replace stack allocated array by heap allocated
- Revert "Efl.Gfx.Path: Change draw of a rounded rect from arc to quadratic"
- Fix typos
- Fix typos - (Part #5)
- eina_array: micro optimize eina_array_push
- docs: Fix typo in elm_config API reference
- gengrid: Fix memory leak
- Fix eina file thread test on Windows
- Windows: fix eina_file_map_new()
- Native Windows: Eina: Resolve bad comparison while using windows strerror_s
- ecore evas wl - cnp - dont segv is data is null
- elementary : fix key unselect unhighlighting issue
- ecore-evas-wayland: Fix minor typo in error message
- elc_naviframe: fix to show pushed item if top item is deleted
- Meson.build cleanup.
- evas_common_privat: improve tiler merge flag
- Protect EINA_(UN)LIKELY with parenthesis around the expr
- Fix typos - (Part #2)
- Fix typos - (Part #3)
- windows: Fix path for file sanitization
- docs: Correct the wrong group name in Edje and Elementary
- evas_text: lazy loading color glyph images in RAM
- benchmark: eina: remove outdated ecore_hash
- edje: Avoid string duplication when possible.
- Revert "edje: Avoid string duplication when possible."
- ibox: Preview Theme Code
- docs: Correct the wrong API group name in Elementary and Change an invalid URL in Evas
- tests: remove popup test.
- Evas: add avif evas loader
- docs: Correct the wrong API group name and typo in Evas, Eet, Eina, Eio and Elementary
- ci: travis: make sure we disable avif loader by default in our jobs
- Revert "Evas: add avif evas loader"
- elm_code: on newline ensure line number visible.
- elm_code: cursor visbility fix.
- Revert "ci: travis: make sure we disable avif loader by default in our jobs"
- edje - fix edje_cc mis-re-iding images in image sets
- Evas: add avif evas loader and saver
- Revert "Revert "ci: travis: make sure we disable avif loader by default in our jobs""
- meson: allow empty values in disabler/bindings array
- ci: travis: force meson version 0.54.3 as install for codecov
- elm: tests: fix typo in test name
- edje_cc : resolve build warning
- edje: Avoid string duplication when possible.
- edje example - dont leak text
- efl_ui_image: remove the unnecessary code in efl_file_unload
- Revert "edje: Avoid string duplication when possible."
- Eina log: flush stderr on mintty-based terminals
- efl: make all _class_get() functions const
- win - x io errors really just have to exit, so call the iorr cb + exit
- elm - genlist+gird - use item specific realize/unrealize not eo callback
- elm - genlist/grid - call realized/unrealized item func cb after eo cb
- edje_textblock: remove duplicated textblock style properties
- evas: font glyphs texture garbage collector
- efreetd - dont confuse log file file handle and log math func
- efreetd - log the cache runs and when they launch, show data, exit
- efreetd - add more logging for what efreetd says to clients
- canvas engine: ++safety code
- build: warn of faulty check version
- ci: do not build tests on ci
- elementry_test: fix allocation size
- evas: remove unused define
- evas - render - dont crash is private data is null
- tests: fix eio tests
- evil: Remove unnecessary delay for evil startup
- evas_textblock : Remove unneeded code
- evas_textblock : return correct value for gfx_filter_get
- eina vpath - fix windows ~username handling to only error when used
- ecore_ipc: remove duplicated condition
- efl_ui_focus_manager: reduce the amount of relying on events
- evas_textblock: remove logically dead code
- elm_config: add slider to configure tooltip_delay
- elua: do not link to cffi, load the module instead
- elua: fix elua_register with 5.1
- elua: add searchpath impl for 5.1 compat
- efl ui image - fix scal to fill - broken with non-rect src images
- evas_gl_core: fix memory leak of EVGL_Resource
- edje_embryo: fix to use parameter of macro ALLOC_COPY_DESC
- edje_embryo: fix null pointer dereference in ALLOC_COPY_DESC
- elm : fix return of gengrid item swap anim to prevent animation memory leak.
- embryo: fix a integer(cell) overflow problem
- edje_embryo: fix memory leak in ALLOC_COPY_DESC
- elput - support libelogind.so.0 as well as libelogind-shared.so.0
- evas: fix a potentional error of null deref
- efl_ui_timepicker: Add null check for efl_text_get
- efl.ui.textbox: Keyboard Enter will add paragraph separator
- evas_textblock: enhance escape character handling
- elm_entry/efl.ui.textbox: use common key shortcuts
- evas_textblock: enhance cursor event submitting during markup_set/text_set
- efl_ui_textbox: preserve changing user color set
- evas_filter: Add null check for instruction
- embryo_cc: ++safty code
- edje_edit: add null check for return of eina_hash_find
- edje_edit: fix a potentional error of null deref
- edje: fix a potentional error of null deref
- evas/engine: fix a potentional error of null deref
- elm_access: fix a potentional error of null deref
- efl_canvas: fix a potentional error of null deref
- edje_entry: fix a potentional error of null deref
- eo: event callback call optimization
- eo: evalulate special counts earlier
- ecore_wl2_subsurf: follow wayland spec for sync
- evil: Remove OS version check
- elm_code: Always render N new lines (buffer).
- elm_code: Selection start (with keyboard) fix.
- ecore input - make an unsupported joystick a warn not an err...
- eo: set idx on exit to 1 not to 0
- evas examples: remove unecessary beta tag.
- edje - calce - really respect max image size for img sets
- ecore_x: Add mode flags new in xrandr 1.2.
- eina: Use INVALID_FILE_ATTRIBUTES instead of magic number
- docs: Correct the wrong API group name in elm_win and elm_glview
- evas_object_textblock: utf8_to_markup support all escapse chars
- Efl Canvas Text : canvas textblock changed not fired for shadow format (style_apply)
- evas_textblock: myanmar script rendering with e vowel.
- eet: Fix 'No OPENSSL_Applink' error message.
- eet test: use eina_file_mkstemp, fix include for vc++
- eina: Close file in unlink test
- eina_test_lock.c: Implement clock_gettime for Windows
- ecore x - go back to vsync thread. this will be more reliable
- ecore - don't do anything with heap between fork and exec
- evas - fix deadlock with sw async rendering calling cbs in post flush
- confs: Add OpenBSD Build Helper Scripts.
- eina - eina_file_close_from - handle coverity theory that closes change
- efl/ecore exe - more paranoia - block sig handlers between fork + exec
- eina file - fix getdents64 on older glibc's and use raw syscall
- ecore_wl2_display: move wl_display_dispatch_pending before condition check.
- release: Update NEWS and bump version for 1.25.0-alpha1 release
- eina file - fd close - handle fbsd 11 syscall abi as it broke in 12
- evas_font_query: add Use parentheses within macros
- ecore con - curl - fix error map to map right enum
- tests - elm slider - fix possile use of uninit var warning
- tests - elm spinner - use proper elm enum - warning fix
- tests - efl ui config - use proepr elm enums for elm apis
- tests - evas vg - use evas enumas not efl ui - warn fix
- tests - evas vg - use proper enuams for evas vg api - warn fix
- exactness - cast enum assign - it is valid due to values. warn fix
- exactness - injecotr - alloc correct buffer size
- evas - gl generic - use correct enums for types
- ecore evas - module - use correct enums
- elm - efl ui win - fix enum handling with explicit casts where needed
- ecore evas - use correct enums for legacy funcstions to remove warns
- evas events - cast enums after checking their values match - fix wrns
- evas - legacy events - cast enums after checking its ok - wrn fix
- evas - sw generic - fix enum passing with casts or correct types fix wrn
- elm image - zoomable - fix enum warnings with valid casts
- elm - efl ui widget - fix enum warnings with valid casts
- elm - focus legacy - fix enum implicit casts with explicit - valid
- elm scroller - fix mapping of block enums - was broken
- elm cnp/dnd - use right evas device enum for seat
- ecore evas, edje - fix use of evas seat device enum
- evas - canavs vg shape - cast enums - valid - checked - fix warnings
- evas - obj main - fix enum warnings with valid casts
- evas - sw generic - fix enum cast warn - valid checked
- evas - image+proxy - fix enum cast warnings with casts - checked
- evas - callbacks - fix casting warnings - checked flags are the same
- elm - flip - fix enum cast warnings - cheked values match
- evas vg - fix warning for enum casts with explicit ones - checked
- evas  - image - fix enum cast warnings - checked
- evas - canvas - fix enum cast warning by removing cast
- edje - fix enum cast warnings - checked values
- elm - main - fix enum cast warnings - checked values
- elm - efl ui text - fix enum return to use right enum
- edje - fix enum cast warning - checked values
- elm access - fix enum cast warning - checked values
- evas device - fix enum cast warnings - checked values
- elm_config: rescale is needed when font or font size is changed
- efl_ui_image: update orientation immediately
- ecore exe - fig signal block to unblock again - some sys dont on exec
- eina - statgen (stat generation) - fix enable api to actually enable
- evas - image cache - fix loading with skip head on to load changed files
- evas load - dont lose skip flg update after async head skip load fix
- elm_theme: more secure check extension as '.edj' and not 'edj'
- eina: Replace remove with rmdir and delete tmpstr
- ecore-file: fix ecore_file_can_exec() on Windows
- Ecore: more precise implementation of ecore_time_get on Windows
- Evil: add gettimeofday()
- release: Update NEWS and bump version for 1.25.0-beta1 release
- ecore_file_can_exec: fix .bat case
- eeze_mount binary: more secure check extension ('.iso' and not 'iso')
- evas_textblock: rainbow flag emoji treated as two clusters(update unibreak to version 4.2)
- evas_textblock: remove style padding from native width and formatted height
- Revert "evas_textblock: rainbow flag emoji treated as two clusters(update unibreak to version 4.2)"
- elm icon/thumb - fix handling of a wider range of video extns
- edje_cc: Fix uninitialized scalar variable
- efl: remove bool and string pointer mixup
- efl: remove unused functions
- eldbus_fake_server: fix string format usage
- benchmarks: specify this outside extern "C"
- efl: resolve warnings over misleading code
- eldbus: use the const macros we have
- efl_ui_focus_manager_calc: fix CID
- build: only find env once
- eina file test: check rmdir result.
- SvgLoader: Defs type nodes are not saved in loader's node list.
- Revert "SvgLoader: Defs type nodes are not saved in loader's node list."
- edje - calc - handle swallow disappearing mid-calc due to callabcks
- docs: Add the missing API page name to elementary and evas
- vg_load_svg: Fix pointing to wrong parent
- ecore drm - increase timeout for flips from 0.05 to 2.0 sec
- ecore x - fix typo in internal backdoor api to get vsync debug
- ecore_evas_x: reply in *any* way with XdndStatus
- elm_cnp: adjust image type to real format
- elm cnp - make matchers const and static too
- elm_gesture_layer: fix crash caused by elm_gesture_layer_cb_del in cb
- release: Update NEWS and bump version for 1.25.0-beta2 release
- evas vg: check object changed as well
- vg_load_svg: Supports case when only rx or ry is declared
- docs: Correct the mismatched group name in Evas.
- Efl.Canvas.Vg.Object: Fix typo
- vg_load_svg: Fix missing loader.gradients update
- vg_load_svg: Add Mask, ClipPath nodees as dummy.
- Efl.Canvas.Vg: When VG container has transparency, ector buffer created synchronously
- release: Update NEWS and bump version for 1.25.0-beta3 release
- vg_load_svg: Support dasharray attribute for stroke
- evas_textblock: prevent textnodes with <ps>/<br> without format node
- evas vg: check object changed as well 2
- evas_main: Fix memory leak.
- evas_outbuf: Fix memory leak.
- call sd_seat_can_tty to determine if ttys should be used, instead of hard-coding seat0
- elm_config: remove useless init|shutdown of ecore_win32 and eocore_cocoa (done in ecore_evas module), and useless wayland variable
- ecore con url - curl is actually optional and only needs to init on use
- elm - collection view - check return and fix coverity wanring
- evas - psd loader - check read values to try make coverity happy
- ethumb - check inoput index from stdin
- ethumb - check size read
- ethumb - fix index range check
- ethumb - fix index range check
- ethumb - fix idx range check
- emotion - check return of eet data write
- eina thread - check return of pthread_attr_init
- evas test - check return of ftell and malloc and handle properly
- eina bench - have default values inside benc dtoa
- ecore - efl exe - check returns of open and dup2 for stdin etc. fds
- ecore - efl thread - dont close invalid < 0 fd's
- elm - toiolbar - use correct theme icon for more item
- Revert "elm - collection view - check return and fix coverity wanring"
- elm_config: don't prepend /tmp/ to file name
- efl: add option to disable lua support
- edge,evas: isolate lua dependency usages


JPTIZ:
...Submitted to the wrong branch. Knew I was forgetting about something :P

PR #471, opened 2020-09-23
Make lua dependency optional

### LINKED ISSUE #464: Make Lua dependency optional
Refer to https://phab.enlightenment.org/T8817

JPTIZ:
This patch adds an option to disable Lua dependency by setting the meson option `lua-interpreter` to `none`. That is:

```console
$ meson builddir -Dlua-interpreter=none
```

This should be enough to disable lua-interpreter.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -261,7 +261,12 @@ ecore_evas_wayland_engine_include_dir = []
 
 evas_static_list = []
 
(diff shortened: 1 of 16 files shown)


JPTIZ:
Should be passing tests now.

PR #472, opened 2020-09-24
win32: Enable dotnet on windows

joaoantoniocardoso:


src/bin/eolian_mono/eolian/mono/type_impl.hh
```diff
@@ -125,7 +125,11 @@ struct visitor_generate
            , {"bool", nullptr, [&] { return replace_base_type(regular, "bool?"); }}
            , {"short", nullptr, [&] { return replace_base_opt_integer<short>(regular); }}
            , {"int", nullptr, [&] { return replace_base_opt_integer<int>(regular); }}
+#ifdef _MSC_VER
+           //, {"long", nullptr, [&] { return replace_base_type(regular, "long?"); }}
+#else
            , {"long", nullptr, [&] { return replace_base_opt_integer<long>(regular); }}
+#endif
```
Coquinho, line 132: Why not an `ifndef`?

src/tests/efl_mono/Array.cs
```diff
@@ -9,14 +9,16 @@ class TestArray
 {
     public static void TestAdd()
     {
+/*
```
Coquinho, line 12: Maybe `#ifnot sys_windows` to disable them only when building for windows?

PR #476, opened 2020-11-13
win32: Replace scripts with native/cross-files definitions

JPTIZ:
By creating so many build scripts, they must be maintained, which
considering the current scripts are batch-scripts it means it will
sometimes be too hard to fix/update them to get too little benefit.

There were two main purposes in having those scripts:
- Ensuring pre-build procedures (setting environment variables and
  running vcvars64) have been done.
- Setting windows-specific project option values, such as disabling
  wayland/x11 support and some Unix-specific EFL modules.

For the first one, after quite a lot of changes in the codebase, there's
only one environment variable to be set: CMAKE_TOOLCHAIN_FILE, which
must then be passed to -Dcmake_args project option to meson. This is now
done by creating a simple native-file with the cmake_args value on it,
so a script to check/define it is above overkill. About vcvars64, this
might be a simple "run vcvars64" (or "start a Develop Console from
Visual Studio") instruction in compile and installation guides.

For the second one, since meson 0.55 and further more on 0.56, there
were some serious updates on cross/native-file definitions, which now
fixes the issue https://github.com/mesonbuild/meson/issues/6783 that was
the one that blocked EFL most from relying solely on native/cross-files.

Given these, there's no good reason left to keep relying on these
scripts, thus simplifying the build procedure a lot and making it more
reliable and customizable, since now we only depend on meson defaults
and not on another language (in this case, batch-script).


platforms/windows-native.txt
```diff
@@ -0,0 +1,35 @@
...
+pulseaudio = 'false'
+harfbuzz = 'false'
+fribidi = 'false'
+fontconfig = 'false'
+edje-sound-and-video = 'false'
+libmount = 'false'
+evas-loaders-disabler = 'gst,pdf,ps,raw,svg,rsvg,xcf,bmp,dds,generic,gif,ico,jp2k,pmaps,psd,tga,tgv,tiff,wbmp,webp,xpm,json'
+opengl = 'none'
+emotion-loaders-disabler = 'gstreamer1,libvlc,xine'
+bindings=
```
joaoantoniocardoso:
```suggestion
bindings = ''
```

platforms/windows-native.txt
```diff
@@ -0,0 +1,35 @@
...
+fribidi = 'false'
+fontconfig = 'false'
+edje-sound-and-video = 'false'
+libmount = 'false'
+evas-loaders-disabler = 'gst,pdf,ps,raw,svg,rsvg,xcf,bmp,dds,generic,gif,ico,jp2k,pmaps,psd,tga,tgv,tiff,wbmp,webp,xpm,json'
+opengl = 'none'
+emotion-loaders-disabler = 'gstreamer1,libvlc,xine'
+bindings = ''
+
+[properties]
```
joaoantoniocardoso:
```suggestion
[built-in options]
```

joaoantoniocardoso (approved):

PR #478, opened 2020-11-13
Devs/jptiz/optional ector

JPTIZ:


meson.build
```diff
@@ -498,6 +498,7 @@ foreach package : subprojects
       endif
     endif
   else
+    set_variable(package_name, declare_dependency())
```
Coquinho: Sorry, but I didn't get why this is needed, could you explain for me?

src/lib/edje/edje_calc.c
```diff
@@ -3262,7 +3271,6 @@ _edje_vector_load_lottie(Edje *ed, Edje_Real_Part *ep, const char *key)
         efl_gfx_frame_controller_frame_set(ep->object, (int)(frame_count * desc->vg.frame));
      }
 }
-#endif
```
Coquinho: Why changing where the `endif` of  `#ifdef BUILD_VG_LOADER_JSON`?

src/lib/elementary/meson.build
```diff
@@ -949,6 +963,10 @@ elementary_src = files([
   'efl_ui_spotlight_animation_manager.c',
 ])
 
+if get_option('ector')
+  elementary_src += 'efl_ui_vg_animation.c'
+endif
```
Coquinho: Does it work? Shouldn't it be:
```suggestion
if get_option('ector')
  elementary_src += files(['efl_ui_vg_animation.c'])
endif
```

src/lib/evas/canvas/efl_canvas_textblock.eo
```diff
@@ -316,7 +316,7 @@ class Efl.Canvas.Textblock extends Efl.Canvas.Object implements Efl.Text,
              Default value is $[1].
              See @Efl.Text_Style.text_underline_height.
 
-           - $gfx_filter: Experimental filter name. See @Efl.Gfx.Filter for more information.
+           - gfx_filter: Experimental filter name. See Efl.Gfx.Filter for more information.
```
Coquinho: What `$` and `@` do here? the first one is to see, `gtx_filter` as a variable a,d the second to set a link to `Efl.Gfx.Filter`?

src/lib/evas/canvas/evas_object_text.c
```diff
@@ -675,10 +681,13 @@ _evas_object_text_pad_get(const Eo *eo_obj, Evas_Text_Data *o, int *l, int *r, i
...
    if (t) *t = 0;
    if (b) *b = 0;
-   if (!o->has_filter)
-     evas_text_style_pad_get(o->cur.style, l, r, t, b);
-   else
+#ifdef HAVE_ECTOR
+   if (o->has_filter)
      efl_gfx_filter_padding_get(eo_obj, l, r, t, b);
+   else
+#else
```
Coquinho: There is no `else` block here Oo

src/lib/evas/canvas/evas_object_textblock.c
```diff
@@ -342,7 +349,9 @@ struct _Efl_Canvas_Textblock_Filter
    Evas_Object          *eo_obj;
    Evas_Public_Data     *evas;
    void                 *dc; /* draw context - no clip, white, no colmul... */
+#ifdef HAVE_ECTOR
    Evas_Filter_Padding   pad;
+#endif
```
Coquinho: I think you don't need an `#ifdef HAVE_ECTOR` inside an `#ifdef HAVE_ECTOR`

meson.build
JPTIZ: At first, some meson.builds were complaining about unknown static_libs, so this was just a shortcut to get it compiling (no preprocessor symbol is defined, so `HAVE_*` won't mislead, etc.).

src/lib/edje/edje_calc.c
JPTIZ: Because it makes not much sense to have vector functions if vg_loader is disabled, I think, but looking at the code I think this was a misunderstanding.

src/lib/evas/canvas/efl_canvas_textblock.eo
JPTIZ: @Coquinho they're referencing other Eo classes. What I do here is prevent the reference, so the doc generator doesn't try to look for them.

src/lib/evas/canvas/evas_object_text.c
JPTIZ: Oh true

JPTIZ:
@Coquinho I addressed some of your comments.

src/lib/evas/canvas/efl_canvas_textblock.eo
Coquinho: Oh, ok.
And why do we need to change docs here?

src/lib/edje/edje_calc.c
Coquinho: For me, it seems like a misunderstanding too, I mean, It was there for a reason, and if it was wrong or not shouldn't be decided at this PR.

src/lib/evas/meson.build
```diff
@@ -134,16 +141,25 @@ evas_include_directories += [
...
 ])
 
 evas_src_opt = [ ]
 
-evas_ext_none_static_deps += dependency('freetype2')
+if get_variable('freetype', false) == false
+  freetype = dependency('freetype2')
+endif
+
+evas_ext_none_static_deps += freetype
```
Coquinho: What was the problem with FreeType?

src/modules/evas/engines/gl_generic/meson.build
```diff
@@ -45,10 +51,12 @@ foreach src : common_engine_src
...
-pub_eo_files = [
-  'evas_ector_gl_buffer.eo',
-  'evas_ector_gl_image_buffer.eo'
-]
+if get_option('ector')
+  pub_eo_files = [
+    'evas_ector_gl_buffer.eo',
+    'evas_ector_gl_image_buffer.eo'
+  ]
+endif
```
Coquinho: Shouldn't `pub_eo_files` be declared empty for the case where there is no ector?
```suggestion
pub_eo_files = []
if get_option('ector')
  pub_eo_files += [
    'evas_ector_gl_buffer.eo',
    'evas_ector_gl_image_buffer.eo'
  ]
endif
```

Or else, the  `foreach` at line61 will yell that there are no `pub_eo_files`, right?

src/modules/evas/engines/software_generic/filters/evas_filter_bump.c
```diff
@@ -103,9 +103,13 @@ _bump_map_cpu_alpha_alpha(Evas_Filter_Command *cmd)
    h = cmd->input->h;
    EINA_SAFETY_ON_FALSE_RETURN_VAL(w > 2 && h > 2, EINA_FALSE);
 
+#define E_READ 0
+#define E_WRITE 0
```
Coquinho: Shouldn't this be defined elsewhere (probably at `src/lib/evas/filters/evas_filter_private.h:310` as an `#else` to the `#ifdef HAVE_ECTOR` at line 308) ? Like this we will overwrite any previous definition, won't we?

src/modules/evas/engines/software_generic/filters/evas_filter_grayscale.c
```diff
@@ -1,4 +1,6 @@
 #include "evas_engine_filter.h"
+#define E_READ 0
+#define E_WRITE 0
```
Coquinho: Same as https://github.com/expertisesolutions/efl/pull/478/files#r523233109

Coquinho (commented):
Do files at `src/lib/evas/software_generic/filters/` need any changes? `src/lib/evas/software_generic/filters/meson.build` has an `subdir_done` in case ector is disabled, and `src/lib/evas/software_generic/meson.build` don't even add `filter` as a subdir in that case.

src/tests/elementary/spec/efl_test_gfx_view.c
```diff
@@ -6,10 +6,12 @@
...
 #include "suite_helpers.h"
 
+#ifdef HAVE_ECTOR
+#else
 /* spec-meta-start
-      {"test-interface":"Efl.Gfx.View",
-       "test-widgets": ["Efl.Ui.Vg_Animation"]}
+      {"test-interface":"Efl.Gfx.View"}
    spec-meta-end */
+#endif
```
Coquinho: Is this as intended? It seems as something went wrong here, or at least it should be a `#ifndef` instead of a `#ifdef`

src/lib/evas/canvas/efl_canvas_textblock.eo
JPTIZ: Because otherwise it won't compile, since if the doc parser doesn't find the references, it gives a compilation error.

src/lib/edje/edje_calc.c
JPTIZ: I'll address that in the next commit.

src/lib/evas/meson.build
JPTIZ: Tbh I find this strange as well. I think I applied an extra `git stash pop` by accident.

src/modules/evas/engines/gl_generic/meson.build
JPTIZ: Yes, or worse: it may get a previously defined `pub_eo_files`.

src/tests/elementary/spec/efl_test_gfx_view.c
JPTIZ: The `#ifdef` here actually makes no changes at all to the comments. This is a leftover of an attempt to catch that behaviour.

src/lib/evas/meson.build
JPTIZ: Oooh right right, now I remember. This is sort of a fix to the `freetype` dependency. First, see the first appearence of `freetype` in `src/lib/ector/meson.build`:

```meson
ector_ext_deps = [triangulator, freetype, draw_without_deps, m]
```

It gets freetype from the static_lib, see the main `/meson.build`:

```meson
  ['ector'            ,['ector']             , false,  true, false,  false,   true,     false,  true, ['eina', 'efl'], ['draw', 'triangulator', 'freetype']],
```

Which defines the `freetype` variable and etc. What I did was to always get the variable when it exists, otherwise get from the system-wide dependency.

src/modules/evas/engines/software_generic/filters/evas_filter_bump.c
JPTIZ: Actually these files aren't even compiled without ector, hm...

src/lib/evas/meson.build
Coquinho: Oh ok, make sense.
But wouldn't be better to do this at the `meson.build` at the root of the project?

Coquinho: Oh, just noticed that `freetype` is declared at `src/static_libs/freetype/meson.build` so maybe there would be even better. What do you think?

JPTIZ: It could be, yes.

src/lib/edje/edje_calc.c
JPTIZ: Done.

src/tests/elementary/elm_test_check.c
```diff
@@ -196,10 +198,12 @@ EFL_START_TEST(elm_test_check_api_call)
 
    get_me_to_those_events(check1);
 
+#ifdef HAVE_ECTOR
    data.value = 0;
    click_object(data.check1);
    ecore_main_loop_iterate();
-   ck_assert_int_eq(data.value, 1);
+   ck_assert_int_eq(data.value, 1); // FIXME No more ifdef have_ector
```
Coquinho: What's the problem here?

JPTIZ: Problem is events' callbacks aren't being called for Eo objects, so that `data.value` is never set.

PR #481, opened 2020-11-17
efl-windows: update to the new way to add path

### LINKED ISSUE #477: Fix deprecated add-path in CI
Recently, [Github Actions deprecated the `add-path` command](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/), as can be already seen in [recent builds](https://github.com/expertisesolutions/efl/actions/runs/361620252). We should replace it.

joaoantoniocardoso:
This PR should fix #477 . 

For reference [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#example-7) is an example from the official documentation.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -33,7 +33,7 @@ jobs:
       if: steps.cache-llvm.outputs.cache-hit != 'true'
       run: choco install llvm
     - name: Add clang-cl to PATH
-      run: echo "::add-path::C:\Program Files\LLVM\bin"
+      run: echo "C:\Program Files\LLVM\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
     - name: Search clang-cl
       run: clang-cl -v
     - name: add python scripts to path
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -119,7 +131,7 @@ jobs:
       run: |
         $python_version = (py -3 --version).replace("Python ", "")
         $python_scripts_path = ($(echo C:\hostedtoolcache\windows\Python\$python_version\x64\Scripts ) -join "")
-        echo "::add-path::$python_scripts_path"
+        run: echo "$python_scripts_path" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```
joaoantoniocardoso:
```suggestion
        echo "$python_scripts_path" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```

JPTIZ (approved):

PR #482, opened 2020-11-19
Add Eo classes workarround for disabiling ector

Coquinho:
This PR define vector-graphics and non vector-grafics eo classes "templates" for:
+ `src/lib/efl/interfaces/efl_text_style.eo`
+ `src/lib/evas/canvas/efl_canvas_group.eo`
+ `src/lib/evas/canvas/efl_canvas_image_internal.eo`
+ `src/lib/evas/canvas/efl_canvas_textblock.eo`


src/bindings/cxx/meson.build
```diff
@@ -63,15 +63,22 @@ foreach lib : cxx_sublibs
     endif
     subdir_file_location = join_paths(file_location, eo_file_subdir)
     foreach cxx_gen_file : cxx_pub_eo_files
+      if (package_name == 'evas' or package_name == 'efl') and cxx_gen_file in pub_eo_files_optional_ector
```
JPTIZ:
```suggestion
      if package_name in ['evas', 'efl'] and cxx_gen_file in pub_eo_files_optional_ector
```

src/bindings/cxx/meson.build
```diff
@@ -63,15 +63,22 @@ foreach lib : cxx_sublibs
...
+        cxx_gen_file_input = cxx_gen_file
+        cxx_gen_file = '@0@'.format(cxx_gen_file)
+        cxx_gen_file = cxx_gen_file.split('/')[-1]
+      else
+        cxx_gen_file_input = join_paths(subdir_file_location, cxx_gen_file)
```
JPTIZ: Since you're already changing this...

```suggestion
        cxx_gen_file_input = subdir_file_location / cxx_gen_file
```

src/lib/efl/interfaces/meson.build
```diff
@@ -1,5 +1,8 @@
 eolian_include_directories += ['-I', join_paths(meson.source_root(), 'src', 'lib')]
 
+# for evas/cavas
```
JPTIZ:
```suggestion
# for evas/canvas
```

src/lib/efl/interfaces/meson.build
```diff
@@ -28,6 +31,38 @@ foreach eo_file : pub_legacy_eo_files
...
+    input : 'efl_text_style_non_vg.eo_template',
+    output : eo_file,
+    copy : true,
+  )
+endif
+
+pub_eo_file_text_style = [meson.current_build_dir() / eo_file]
+pub_eo_file_target += custom_target('eolian_gen_' + eo_file,
+  input : meson.current_build_dir() / eo_file,
+  output : [eo_file + '.h'],
```
JPTIZ: why an array?

src/lib/efl/interfaces/meson.build
```diff
@@ -28,6 +31,38 @@ foreach eo_file : pub_legacy_eo_files
...
+  )
+else
+  configure_file(
+    input : 'efl_text_style_non_vg.eo_template',
+    output : eo_file,
+    copy : true,
+  )
+endif
+
+pub_eo_file_text_style = [meson.current_build_dir() / eo_file]
```
JPTIZ, line 47: why an array?²

(I know you add it into an array later on, but don't worry, it doesn't have to be one: `array + element` appends the element into the array)

src/lib/evas/canvas/efl_canvas_textblock_non_vg.eo_template
```diff
@@ -316,6 +316,8 @@ class Efl.Canvas.Textblock extends Efl.Canvas.Object implements Efl.Text,
              Default value is $[1].
              See @Efl.Text_Style.text_underline_height.
 
+           - gfx_filter: Experimental filter name. See Efl.Gfx.Filter for more information.
```
JPTIZ: If you're getting vg back, you can readd the doc references here:

```suggestion
           - $gfx_filter: Experimental filter name. See @Efl.Gfx.Filter for more information.
```

src/lib/evas/canvas/meson.build
```diff
@@ -1,7 +1,49 @@
+pub_eo_files_optional_ector_ = [
+  ['efl_canvas_group.eo', 'efl_canvas_group_vg.eo_template', 'efl_canvas_group_non_vg.eo_template'],
+  ['efl_canvas_image_internal.eo', 'efl_canvas_image_internal_vg.eo_template', 'efl_canvas_image_internal_non_vg.eo_template'],
+  ['efl_canvas_textblock.eo', 'efl_canvas_textblock_vg.eo_template', 'efl_canvas_textblock_non_vg.eo_template'],
```
JPTIZ: Not too much, but...

```suggestion
  ['efl_canvas_group.eo',
   'efl_canvas_group_vg.eo_template',
   'efl_canvas_group_non_vg.eo_template'],
  ['efl_canvas_image_internal.eo',
   'efl_canvas_image_internal_vg.eo_template',
   'efl_canvas_image_internal_non_vg.eo_template'],
  ['efl_canvas_textblock.eo',
   'efl_canvas_textblock_vg.eo_template',
   'efl_canvas_textblock_non_vg.eo_template'],
```

We could also change it into a dict (for better semantics) or just store the prefix and add `_vg` and `_non_vg` + `.eo` and `.eo_template` inside the `foreach` loop.

src/lib/evas/canvas/meson.build
```diff
@@ -70,7 +110,8 @@ foreach eo_file : priv_eo_files
                            '-gchd', '@INPUT@'])
 endforeach
 
-evas_canvas_eo_files = pub_eo_files
+
```
JPTIZ: But why

src/lib/evas/canvas/meson.build
```diff
@@ -1,7 +1,49 @@
+pub_eo_files_optional_ector_ = [
```
JPTIZ: For temporary/internal variables, I prefer to _prefix_ with `_` instead of suffixing, since it is more unlikely that someone should be using a variable that starts with `_`.

```suggestion
_pub_eo_files_optional_ector = [
```

src/lib/evas/canvas/meson.build
```diff
@@ -1,7 +1,49 @@
...
+  endif
+  eo_file = eo_file[0]
+
+  configure_file(
+    input : in_eo,
+    output : eo_file,
+    copy: true,
+  )
+
+  priv_eo_file_target += custom_target('eolian_gen_' + eo_file,
```
JPTIZ, line 22: Where is this variable initialized?

src/tests/eolian/eolian_static.c
```diff
@@ -10,6 +10,7 @@ EFL_START_TEST(eolian_static_check)
 {
    Eolian_State *eos = eolian_state_new();
    fail_if(!eolian_state_directory_add(eos, EO_SRC_DIR));
+   fail_if(!eolian_state_directory_add(eos, EO_BUILD_DIR));
```
JPTIZ, line 13: Why do we need this? (And what does this test-case tests exactly?)

src/bindings/cxx/meson.build
Coquinho: ohhh, good point

src/bindings/cxx/meson.build
Coquinho: sure.

src/lib/evas/canvas/efl_canvas_textblock_non_vg.eo_template
Coquinho: This one is the non_vg version, that's why it is like this.

JPTIZ: I read the `vg` in the filename and assumed it was for vg haha.

Well, since this is non-vg, you can remove that entire line so, since the attribute won't exist.

Coquinho: Actually, what do you think about just removing this line?

JPTIZ: That's exactly what I just said.

Coquinho: Sorry, the page needed aa refresh here

src/lib/evas/canvas/meson.build
Coquinho: `src/lib/eo/meson.build:28:priv_eo_file_target = []`

and elementary rewrite it:
`src/lib/elementary/meson.build:250:priv_eo_file_target = []`

JPTIZ: Hm, do we want to append? How does it work/how is it processed?

src/lib/evas/canvas/meson.build
Coquinho: I've reworked this, what do you think about it

src/lib/efl/interfaces/meson.build
Coquinho: I've tried this, but it yielded error:
```
ERROR: Invalid use of addition: can only concatenate list (not "str") to list
```

src/lib/evas/canvas/meson.build
Coquinho: I actually don't know how/why this works/is processed, here I'm just mimicking [this](https://github.com/expertisesolutions/efl/pull/482/files?file-filters%5B%5D=.build&file-filters%5B%5D=.c&file-filters%5B%5D=.cc&file-filters%5B%5D=.eo_non_vg#diff-724e8a6f046442f85b7751564ca86b86dd22fb35b9d5e56b9c3b345b6c14fbb8L62), as it was done before this patch. So, just maintaining the same semantics.

src/tests/eolian/eolian_static.c
Coquinho: This kinda loads all the `.eo`s at that directory, the next line compiles them all. As now there are some `eo`s classes at the build dir, we need to include them as well to address all dependencies.

